PR #193, opened 2015-01-05
Add latin extended character sets for Polish characters

dkorsak:


lenamtl:
Thanks

almasaeed2010:
I'd keep that as an option since it will increase the page load time for users who don't need it. Thanks!

kressly:
Is the latin not the character sets that enable French language for example ?

PR #200, opened 2015-01-09
Fix visual problem with button en header boxes

Chofoteddy:
Visual problem is corrected with the buttons and dark text.


almasaeed2010:
This issue has been fixed in AdminLTE 2.0.

PR #213, opened 2015-01-18
add var icheck-minimal-path

virhi:
Hello

This commit just give us the possibilty to choose the path of iCheck minimal  images.


almasaeed2010:
iCheck is not mandatory anymore and has been removed from the less files. Thank you for the contribution!

PR #247, opened 2015-02-21
Fix typo + clean spaces

shaialon:


kressly:
What does ths do ? Should we download the master from Github or from almasaeed ?

PR #259, opened 2015-02-26
make sure that toggle sidebar class works properly

yajra:
This PR will:
- make sure that toggle sidebar class works properly
- fix issue when printing and sidebar-collapse still occupies the space
- fix content-wrapper height when expanding treeview menu.
   - this pushes the footer in relation to sidebar's new height
- add initial body class for sidebar toggle function to work well


almasaeed2010:
Did you make sure that this works on small devices too? The class sidebar-open only affects small screens. We implemented it this way so the sidebar is open by default on large screens and collapsed by default on small screens.

yajra:
Have not tested on actual small device but have tested it by resizing the browser and refreshing the page multiple times. Having these issues on my project and these patches fixes it.

yajra:
And yes, sidebar is open for large screen and collapse on small screens.

almasaeed2010:
I'll download your fork and test it on small devices. We appreciate the contribution!

almasaeed2010:
I noticed a small issue. The sidebar on small screens doesn't open from the first click. I have to click twice to open the sidebar. After that, the toggle button works fine. Why do you think that is?

almasaeed2010:
I think this is what we should do to make sure that it functions fine on both devices and fixes the print issues. In your commit at line 252 you added ```.addClass('sidebar-collapse');``` which is what we need. However, we should keep the toggleClass function since there is a handler now for small devices that will make sure that sidebar-collapse is set in the body tag.

almasaeed2010:
Let me know what you come up with. Currently I have some other work so I will attend this issue later. Thank you!

yajra:
I think the problem was there was no initial body class of sidebar-open/collapse which causes the issue. Will work on this and let you know.

yajra:
"The sidebar on small screens doesn't open from the first click" fixed on https://github.com/yajra/AdminLTE/commit/ecf2e4a5312e3f491456562b3d1ddd6df99b259c. Kindly check again. Thanks!

almasaeed2010:
Hi again. I tested this and it works fine. However, I don't see why is this necessary. If you look at the CSS file, you will see that sidebar-open overrides sidebar-collapse on small screen automatically. 

You mentioned earlier that this fixes printing issues, how does it do that? The CSS has a ```@media print``` queries that overrides any margining or padding in the layout so ```sidebar-collapse``` and ```sidebar-open``` don't take effect when printing. 

We are still experiencing issues with printing. Especially from chrome, which for some reason is ignoring the print queries. I am still not sure what is causing this issue.

yajra:
The problem was on cases when sidebar-collapse and sidebar-open are both added on the body at the same time due to lack of initial class.

For printing on V1 when sidebar is collapse, the whole content is printable while on V2, the sidebar still occupies the printing space.

almasaeed2010:
I see the problem now. After a few tests, I noticed the sidebar hides only if the sidebar-collapse class is applied and the sidebar-open is not. The goal that we want to achieve is to have the printing work weather these classes are there or not. I think the way to do that is via CSS. I am currently making this my priority since the template promises an enhanced printing experience.

almasaeed2010:
Hello again, 
So sorry for the delay. Could you please create a pr for the second solution? I fixed some printing issues using CSS and now I think we can implement your code without affecting the normal behavior. 

Thank you!

yajra:
No problem. Will create a new PR for each issue later. Closing this now since it's not sync with master branch. Thanks!

yajra:
thanks and sorry, quite busy atm and was not able to create the new PR! 

seems like this were fixed on v2.0.3. :beers:

almasaeed2010:
No problem, I added few things in the css so I decided to implement the fix myself. Thank you for the contribution!

PR #276, opened 2015-03-04
Fixed .js filename in bower.json

albelop:
Fixed bower.json to match the javascript file on /dist/js


almasaeed2010:
Thank you. I didn't realize I wrote it wrong!

PR #298, opened 2015-03-20
As a user I should be able to add values by tag

laudeco:
Hello,

As a user I want to be able to add values by tag. For that I use this lib : Bootstrap Tags Input (https://github.com/timschlechter/bootstrap-tagsinput).

I add it in advanced form elements by 2 basics example.
I don't know your feelings about it.

Thanks
L.


almasaeed2010:
Hi,

I will take a look at it soon and get back to you. Thanks!

fschaeffer:
big +1, this is one of of the missing features to this awesome theme. I included chosen to my own implementation but the bootstrap-tagsinput looks way better and promising.

YasserHassan:
I suggested enhancement #306 for [Select2](http://select2.github.io) which not only supports tagging but also many more additional features.

fschaeffer:
Yep this would also be a choice.... At least I think one of the following tag solutions should be included:

* [Select2](http://select2.github.io) - the de-facto-standard
* [Chosen](http://harvesthq.github.io/chosen/) - quite good for tags, jQueryUI needs rebranding for bs3
* [TagsInput](https://github.com/timschlechter/bootstrap-tagsinput) - already beautiful styled and matching the overall bs3 experience

If you wanted to include a full blown solution which also supports multiple selects, language support, theme support and RTL we'd go for Select2

almasaeed2010:
I think since we are implementing select2, there won't be a need for another similar script. I greatly appreciate the contribution and hope to see more in the future. Thank you!

PR #313, opened 2015-03-29
Add rtl support css files

airani:
Hi,
I'm create RTL (Right to Left) version of AdminLTE and hope so you add this feature to the AdminLTE 
Thank you :smiley:


almasaeed2010:
Thank you for the contribution. I will look into it soon.

almasaeed2010:
I am afraid you did not use less files which is a main issue here since AdminLTE is customized through them. I am working on an RTL version that uses less. However, I'd like to add a link in the documentation to your version until I add the rtl less files.

airani:
Don't afraid Abdullah i'm here :smiley: I should work on less rtl files but until you can use this css files.

almasaeed2010:
@airani are you using a script that generates the RTL files?

YasserHassan:
By LTR, do you mean supporting left-to-right text direction inside the different page elements or mirroring the whole page, making the menu sidebar and logo to the right, ...etc?

airani:
@almasaeed2010 actually i wrote a script for myself to create rtl version from css files. why you ask this question?

almasaeed2010:
@YasserHassan RTL supports right to left languages such as Arabic and Persian. It would require to have the layout somewhat flipped. That is, the main sidebar and logo on the right and RTL text direction.

@airani that's interesting. I ask because a rule in software engineering suggests that if you can automate a process to achieve the same result, you should do so. It would be much easier to add a grunt task, for example, to generate the RTL version instead of having to recreate each element. However, there are few considerations to keep in mind. For instance, how will we deal with directional classes such as ```.pull-right``` which should make an element float to the right. It wouldn't make sense to make ```pull-right``` pulls an element to the left or ```text-left``` to set text alignment to the right. While automation might be great, will the result of all elements make sense to the end-user?

otamimi:
hello, thank you @almasaeed2010 for a great contribution to the community, and @airani for the rtl.
how soon should we expect the RTL version to be released? I am in disperate need for it

YasserHassan:
I suggest investigating approaches that rely on CSS only with no to minimum HTML change such as mirroring via transformation and setting RTL on the document level affecting everything underneath to avoid the need to maintain two different HTML layouts.

Regarding scripts, I suggest taking a look at the following:
- [CSSJanus](http://code.google.com/p/cssjanus/)
- [RTL/LTR for Less](http://github.com/Pomax/node-rtltr-for-less)

And also this blog:
- [Right to left localization with pure CSS selectors](http://alicoding.com/right-to-left-localization-with-pure-css-selectors/)

airani:
@almasaeed2010 When we want to create rtl version from a ltr style we are inevitably to inverse all directions, for example in `.pull-right` we should have `float: left;` but if we create from base a theme in rtl style we can use `.pull-right` with `float: right;` but for ltr style again we need inverse all directions.

almasaeed2010:
#### Update
I looked into automating the conversion from LTR to RTL. I have previously used a plugin called [R2](https://www.npmjs.com/package/grunt-r2). Unfortunately, it does not support LESS, so the users won't be able to customize the LESS files easily. They will have to use the Grunt file to manually trigger the R2 plugin. The next release (Beta 1) should provide support to RTL if all goes well.

almasaeed2010:
Okay I give up. I tried several times with several different scripts to create the rtl version. Many issues arise. I looked into your version and it has many issues too. The sidebar does not function properly. Try clicking the toggle button and you will see the sidebar dispositioned. Instead of creating separate files for the rtl version, you have overridden the original file which is causing a massive overhead. This can be avoided by just using the rtl files you create. I appreciate your effort but I am gonna have to cancel the rtl support. I apologize for any inconvenience and thanks.

lxsameer:
As I mentioned in #97 you can use `bi-app-less` to simultaneously develop AdminLTE for both RTL and LTR layouts. It's not a hard thing to do at all, I'm using my own RTL fork in production for several month now, and it's working like a charm.

YasserHassan:
@lxsameer I agree that RTL support can increase the popularity of AdminLTE. Please update us with the results of converting the new version of AdminLTE and whether you can overcome the issues @almasaeed2010 mentioned above in an automated manner.

lxsameer:
@YasserHassan sure, why not

almasaeed2010:
@lxsameer the plugin you suggested doesn't support the ```transform``` property which is what's causing the main issue. I appreciate your suggestion.

lxsameer:
@almasaeed2010 I believe that adding support for `transform` wouldn't be so hard.

almasaeed2010:
I don't think I have the time to put so much effort into this right now. However, if I get a working example, I might go from there and continue development for the RTL version. My idea would be to completely separate the two versions in terms of CSS files. That is so we don't cause unnecessary load on the main version. Also, I'd like to point out the ```bi-app-less``` script is no longer maintained by its creator. Perhaps we can find a better script that does a similar job. 

I still can't say for sure if I will attempt this soon though since it caused a lot of issues before.

lxsameer:
I have plans to update my fork with the main stream AdminLTE and the do these stuff in the new version too. As I mentioned earlier I have RTL enabled AdminLTE on production for a while now. So 
I made a pull request in the future with the RTL support.

almasaeed2010:
I would appreciate the help and hope we get to make it happen.

lxsameer:
Me too :wink:

YasserHassan:
@lxsameer my understanding is that you use a custom version of `bi-app-less`. This is not a safe approach to stick to as the project itself seems abandoned (last update was 2 years ago) and you'd need to keep updating your customization yourself, if not the script itself if/when it breaks. My opinion is that this tool is safe for a case-by-case use but not as an integral part of AdminLTE. It's better to stick to a maintained tool that provides all what AdminLTE needs out of the box. Otherwise it's better to keep the RTL version as a fork, not part of the main project.

lxsameer:
@YasserHassan Dude with all due respect, I'm not agree with you. Each tool have a goal. And obviously it would be less active after hitting its goal specially in case of such a small tool. 
Of Course it would be better to have a active tool for this but it's not a rocket science, It's a simple less flipper. I don't have any problem with forking, that's what I did until now and I can continue with that too.

YasserHassan:
@lxsameer if the project is active then the transformation issue @almasaeed2010 mentioned could've been reported there and expected to be fixed/enhanced in a decent time whether by the original developer or by a PR from you for example, right? Instead you needed to fork it to enhance/fix it. Then what about the next issue/need discovered? And the next?... The tool is not by its size, once it becomes part of your infrastructure, you expect it to be reliable enough to forget about it and focus on your work. This tool would be responsible for generating a whole RTL edition of AdminLTE so - despite of its small size - if a single line of code breaks, that RTL version could be broken. That's why I recommend again keeping the RTL version as a separate fork as long as there's no long-running maintainable tool to generate it reliably out of the box.

lxsameer:
@YasserHassan To be honest I don't care about the upstream **AdminLTE**. As long as I can have my RTL version working I would be happy. I don't have any problem with continue as a fokr. I just wanted to help so if you think this isn't helpful don't worry just continue as you want.

YasserHassan:
@lxsameer it's not about being helpful, any effort is helpful or at least should be appreciated. It's the matter of interest and different time perspectives. Considering the health and continuity of a project like this calls for avoiding individual and ad hoc solutions. As you mentioned yourself:
>To be honest I don't care about the upstream AdminLTE. As long as I can have my RTL version working I would be happy.

While the premise of my opinion is keeping AdminLTE stable and usable for everyone - including myself - for a long time. At the end, benefit of this discussion is examining the different ideas in order to help @almasaeed2010 decide what's best for his project.

lxsameer:
@YasserHassan I think this argument is vague. At the end of the day I'll do what ever I need to extend my fork as I need and if you guys liked it you can merge it, if not good luck on your path and good job.

PR #316, opened 2015-04-01
[WIP] Jade implementation

Soullivaneuh:
This PR is a dev env proposition.

Pages template are written with [jade-lang](http://jade-lang.com/) and compiled into html with grunt.

### TODO

 - [ ] Rewrite all pages to jade-lang
 - [ ] Add some documentation

### Why change the template system?

Initially, I wanted to make a little PR that modify the pages layout (e.g. header, content).

I cloned my fork, open the `pages` folder and saw my pain: All html files are completely duplicate with no layout refactor!

So you would say "Normal, it's html!". You are right, it's because I propose to change it to a little more advanced template system.

Using a template system with factored code will render this project more accessible for dev contributions.

### Why Jade?

Good question. I was looking for a simple html template system just to separate layout logic from content.

After taking a tour on [template-chooser](http://garann.github.io/template-chooser/), I was seduced by Jade.

BTW, Jade is used by bootstrap too: https://github.com/twbs/bootstrap/blob/master/Gruntfile.js#L320-L333

After all, it's not an imposition. If somebody has a better proposition, feel free to advice it on this PR! :+1: 

### How it works?

To check my Jade implementation, do the following commands.

```sh
git clone https://github.com/Soullivaneuh/AdminLTE.git
cd AdminLTE
git checkout jade
grunt jade watch
```

And go to the `index.html` page.

If you want to edit a template or layout, just edit the related `.jade` and grunt will compile it for overwrite  html files.

In order be the most strait with the current directory structure, all the jade files are on `pages/_jade` and grunt output them on root directory.

I don't think is the best and proper solution. Maybe we should do something like this:
```
pages/src/**/*.jade
pages/dist/**/*.html
```
But we'll have some problem with css relative links. Maybe a cssrewrite system will be required.

Or maybe a global directory restructure like this:
```
src/less/**/*.less
src/views/**/*.jade
src/scripts/**/*.js

src/dist/css/*.css
src/dist/js/*.js
src/dist/pages/*.html
src/dist/index.html
src/dist/index2.html
```

### Can we do more with this?

Of course yes! I make this proposition just for dev env in order to make pages modification easier.

But we can make it better with an official jade-lang support by adding some mixins for our AdminLTE widgets. Based on this example for bootstrap: http://webdesign.tutsplus.com/tutorials/baking-bootstrap-snippets-with-jade--cms-22798

But I think it's good for another PR. One feature at a time. :+1: 

### Your choice!

As I said, it's a proposition. I'll be glad see it accepted. :)

This PR is not complete, I just re-factored the layout and index.html to give you an example.

If we are agree with this proposition, I'll finish it. :smile: 

Thanks for reading.

Regards.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,13 +4,24 @@ module.exports = function (grunt) {
   grunt.loadNpmTasks('grunt-contrib-less');
   grunt.loadNpmTasks('grunt-contrib-watch');
   grunt.loadNpmTasks('grunt-contrib-uglify');
+  grunt.loadNpmTasks('grunt-contrib-jade');
   grunt.loadNpmTasks('grunt-includes');
-  
+
   grunt.initConfig({
     watch: {
-      // if any .less file changes in directory "build/less/" run the "less"-task.
-      files: ["build/less/*.less", "build/less/skins/*.less", "dist/js/app.js"],
-      tasks: ["less", "uglify"]
+      // Redo specific build on file modification
+      css: {
+        files: ["build/less/*.less", "build/less/skins/*.less"],
+        tasks: ["less"]
+      },
+      scripts: {
+        files: "dist/js/app.js",
+        tasks: ["uglify"]
+      },
+      html: {
+        files: ["pages/_jade/**/*.jade"],
+        tasks: ["jade"]
+      }
     },
     // "less"-task configuration
     //this task will compile all less files upon saving to create both AdminLTE.css and AdminLTE.min.css
@@ -36,7 +47,7 @@ module.exports = function (grunt) {
       //production compresses version
       production: {
         options: {
-          //Whether to compress or not          
+          //Whether to compress or not
           compress: true
         },
         files: {
@@ -64,11 +75,30 @@ module.exports = function (grunt) {
         }
       }
     },
+    jade: {
+      compile: {
+        options: {
+          pretty: true,
+          data: {
+            debug: false
+          }
+        },
+        files: [
+          {
+            cwd: "pages/_jade",
+            src: ["*/*.jade", "index*.jade"],
+            dest: "",
+            expand: true,
+            ext: ".html"
+          }
+        ]
+      }
+    },
     //Build the documentation files
     includes: {
       build: {
-        src: ['*.html'], // Source files 
-        dest: 'documentation/', // Destination directory 
+        src: ['*.html'], // Source files
+        dest: 'documentation/', // Destination directory
         flatten: true,
         cwd: 'documentation/build',
         options: {
--- a/index.html
+++ b/index.html
@@ -2,276 +2,146 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <title>AdminLTE 2 | Dashboard</title>
-    <meta content='width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no' name='viewport'>
-    <!-- Bootstrap 3.3.2 -->
-    <link href="bootstrap/css/bootstrap.min.css" rel="stylesheet" type="text/css" />    
-    <!-- FontAwesome 4.3.0 -->
-    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css" rel="stylesheet" type="text/css" />
-    <!-- Ionicons 2.0.0 -->
-    <link href="http://code.ionicframework.com/ionicons/2.0.1/css/ionicons.min.css" rel="stylesheet" type="text/css" />    
-    <!-- Theme style -->
-    <link href="dist/css/AdminLTE.min.css" rel="stylesheet" type="text/css" />
-    <!-- AdminLTE Skins. Choose a skin from the css/skins 
-         folder instead of downloading all of them to reduce the load. -->
-    <link href="dist/css/skins/_all-skins.min.css" rel="stylesheet" type="text/css" />
-    <!-- iCheck -->
-    <link href="plugins/iCheck/flat/blue.css" rel="stylesheet" type="text/css" />
-    <!-- Morris chart -->
-    <link href="plugins/morris/morris.css" rel="stylesheet" type="text/css" />
-    <!-- jvectormap -->
-    <link href="plugins/jvectormap/jquery-jvectormap-1.2.2.css" rel="stylesheet" type="text/css" />
-    <!-- Date Picker -->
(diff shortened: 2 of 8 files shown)


almasaeed2010:
This is very interesting and a well-written proposition! I have a few concerns and comments but I will get back to you when I get some time (probably by the weekend) to give it enough thought. I am currently dealing with school related projects and I really would like to look into this a bit further. I appreciate your patience.

YasserHassan:
The evaluation of this proposition needs to consider the fine-turning and accurate adjustment of the resulting HTML. Will it be possible to produce each and every page directly from Jade without needing to touch the resulting HTML? Otherwise it will have a negative impact on maintainability.

Soullivaneuh:
@almasaeed2010 OK I'm waiting your feedback. Thanks.

@YasserHassan Not sure to understand your question.

> Will it be possible to produce each and every page directly from Jade without needing to touch the resulting HTML?

The Gruntfile is already configured for it. If you have the following directory structure:

```
index.jade
index2.jade
pages/widgets.jade
pages/calendar.jade
```

It will produce:

```
index.html
index2.html
pages/widgets.html
pages/calendar.html
```

But obviously, any modification must be set on jade file, not on html file.

Did I correctly answer your question?

Soullivaneuh:
I missing one command line to test my PR. Just updated the main description.

YasserHassan:
@Soullivaneuh what I meant is that sometimes you need to make fine adjustments to the HTML to achieve the design you want. For example, line breaks inside the HTML (clicking enter) sometimes produce undesirable effects. Currently, to fix such a problem, you need to edit the HTML source and remove that line break. And this is only an example.

If you can not reach such level of control with the template engine, the pipeline could be broken where manual intervention is required.

Soullivaneuh:
I'm sorry but why your adjustments couldn't be make on jade files?

Just run `grunt watch` to have compiled html instantly when you do modification on jade files.

Or I (again) didn't understand. In this case, can you give me a real use case of your explanation?

Thanks.

YasserHassan:
@Soullivaneuh the following HTML fragments are supposed to be the same.

Case 1:
```
<div>
    <div class="a"></div>
    <div class="b"></div>
</div>
```
Case 2:
```
<div>
    <div class="a"></div><div class="b"></div>
</div>
```
But in some rare situation, they are rendered differently and you need to make sure that your HTML is exactly as in Case 2 not 1. Again this is only an example.

Also @almasaeed2010 uses helpful comments to point to the block ends such as `<!-- start message -->`, `<!-- end message -->`, are they easily reproducible using Jade (or any other template engine), or even easier to produce.

You wrote:
>All html files are completely duplicate with no layout refactor!

This should not be a problem from the web developer's point of view. The HTML version is the most common format everyone can start form. From there you take the `starter.html` page as mentioned in the documentation and extract the shared layout to reuse. If the template was done in a PHP for example things would be different, but since this template is for everyone, HTML is the way to go. There could be a benefit from @almasaeed2010's side though as he wouldn't need to wok on many HTML pages with duplicate code, reducing the work required maintain the pages. However, there are solutions to achieve this without a template. For instance, I imagine using HTML inside PHP with shared layout for navbar, sidebar, ...etc, and producing the site as multiple HTML pages from it somehow.

Using Jade or another alternative could be a good idea but we need to make sure of the implications of such decision on the future of this project.

Soullivaneuh:
Ok I got it.

For your use cases, I think Jade can handle needed space. For example:

No space:
```
i.fa.fa-icon
| Dashboard v1
```

One space separation:
```
i.fa.fa-icon
|  Dashboard v1
```

But the best to know if this works properly is to try it entirely, that I will do if @almasaeed2010 is agree. :+1: 

> However, there are solutions to achieve this without a template.

As I said, I'm open to hear proposition and I'm curious to head yours! ;)

> From there you take the starter.html page as mentioned in the documentation and extract the shared layout to reuse. If the template was done in a PHP for example things would be different, but since this template is for everyone, HTML is the way to go.

I agree, btw pages are here for demo and documentation.

But it's especially the goal of this PR: permit a more easy to edit system if a contributor has to make some modifications on the layout system to work with.

In my case, I wanted to make a PR that fix some Twitter Bootstrap code rules issue. So template modification on layout is needed. With the actual case, I have to modify n `.html` template. With the template system, only 1. :+1:

almasaeed2010:
The reason I want a template system is to seperate the navigation files such as, sidebar, main-header and right sidebar from the content. I hate having to add a link manually to each and every page that has a sidebar when I create a new features. The template system will eliminate the need to do so over and over again. We will still have the html files, of course, commented and properly organized as we are used to. 

Regarding the structure I believe we should add our template pages in the build directory within a special directory: 
```
Build/
--------Less/
--------Bootstrap-less/
--------Html/
          ------- the template pages will be added here and the compiled version would be where the current html files are.
```
This way no one not in the development team has to use the template system and our tasks get less hideous.

YasserHassan:
I thought @almasaeed2010 has figured out a way for reusability with components like the menu and sidebar but since it's not, I'm more biased for adopting templates. This alone is a strong reason.

However, I prefer that this step is not done immediately. This is an important decision that calls for good evaluation and selection. I suggest to postpone this move till after v2.1.0 until it's discussed and researched thoroughly.

Soullivaneuh:
> I hate having to add a link manually to each and every page that has a sidebar when I create a new feature.

Not sure to understand. It's exactly why I make this proposition. What is the problem so?

@YasserHassan Sure. As I said, it's a proposition and my PR is not finished at all. ;)

almasaeed2010:
@Soullivaneuh there is no issue. I was adding to your reply as to how we can benefit from Jade.

Soullivaneuh:
Ha ha ok! Was a little confused sorry. ;)

almasaeed2010:
I decided to not use Jade until version 3.0 or perhaps with another open source template. I greatly appreciate your effort and suggestion!

Soullivaneuh:
> I decided to not use Jade until version 3.0

So why close this PR? Could be still opened and merge only for 3.0. ;)

> or perhaps with another open source template.

For 3.0 or 2.x? Which another template system would you like to use?

almasaeed2010:
Git would not allow merge when there are so many file changes. Thus rendering this PR not useful in the future. 
The other template I mentioned was creating a completely different control panel theme with the release of Bootstrap 4. Please understand that I do like the idea but I don't think it is appropriate at this point of development, nor will it be until a major change in the template structure occurs. Thank you.

Soullivaneuh:
Ok I understand that. I'm staying in touch, thanks!

PR #320, opened 2015-04-03
Allow configure image path for boxed-bg.jpg

Elbandi:
If adminlte is included to other projects or merge with other css files, the builded all.css is not relative "close" to image directory (like laravel public/img and public/build/css elixir directory)

This patch allow configure the image path, where the boxed-bg.jpg is copied after build.
Same feature in bootstrap:

https://github.com/twbs/bootstrap/blob/master/less/variables.less#L79

https://github.com/twbs/bootstrap/blob/master/less/glyphicons.less#L13


almasaeed2010:
This is a good idea. I think we should put the full path including the image name into the variable. This way the user gets the ability to change the image using only the variables too. Thanks!

YasserHassan:
It would be good to do the same in the HTML too for all types of assets like JS and CSS if you decided to go with a template engine like Jade which can help make the directory structure very flexible and adopt to the different developers' preferences.

PR #325, opened 2015-04-06
Google web font url fixed

dr1v3:
Fixed link to the font. Page no longer tries to load the font from the hard disk and there is no delay in 20 seconds.

### DIFF
--- a/dist/css/AdminLTE.css
+++ b/dist/css/AdminLTE.css
@@ -1,4 +1,4 @@
-@import url(//fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,600,700,300italic,400italic,600italic);
+@import url(http://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,600,700,300italic,400italic,600italic);
 /*!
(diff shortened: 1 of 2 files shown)


almasaeed2010:
Hello,
I see the issue. This wouldn't be an issue when deployed to servers (or localhost). However, I'd also like it to work without servers and work on https protocol. In order for that to work, we need the font to load via https which will work on all sites whether they use SSL or not. Also, this must be done on the LESS files not the CSS files. I appreciate your contribution. I will apply these changes and recompile the CSS files soon.

kressly:
I had faced some challenges with google font while working localy. It was keeping long to load and was slowing my application . So in Adminlte.css I commented out the font line. 
@almasaeed2010 I hope if you compile it with the less file the page loadin will not be slow if the internet I slow. Bcose not everybody is using the google font so I suggest it shouldn't be compulsory.

almasaeed2010:
The protocol won't fix loading speed issues. Of course, the template provides a fallback font in case you don't want to use the font. 

Fixed in master.

PR #340, opened 2015-04-15
Update FullCalender to v2.3.1

ifaniqbal:


almasaeed2010:
This would interfere with the current style of the fullcalendar that is used with AdminLTE. I will look into this further at a later time. Thanks.

PR #361, opened 2015-04-22
Fix Bootstrap version number on starter page

suvjunmd:
Maybe it would be better to remove the version number from the comments, so there would be no need to update it every time.


almasaeed2010:
I fixed this on all pages. Thanks for reporting this!

PR #368, opened 2015-04-26
Removed closing tag from input elements

suvjunmd:
Input is a void element which can only have a start tag, end tags must not be specified.

http://www.w3.org/TR/html5/syntax.html#void-elements

### DIFF
--- a/documentation/build/include/components.html
+++ b/documentation/build/include/components.html
@@ -906,7 +906,7 @@
           <h3 class="box-title">Input</h3>
           <div class="box-tools pull-right">
             <div class="has-feedback">
-              <input type="text" class="form-control input-sm" placeholder="Search..."/>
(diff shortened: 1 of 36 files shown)


almasaeed2010:
Hello,
Input elements and multiple others such as img and link allow for an optional "/" to indicate end of tag. If I merge your request, the code would be inconsistent since you didn't change img, link, br and other tags. Also, I am working on the premium collection which uses this syntax. Thanks for the work though!

suvjunmd:
I appreciate your effort to keep the code consistent, but you sholud be aware that there are quite a few places where the input element is used without a closing tag:

input

almasaeed2010:
I'll take a look at that when I get a chance. Thank you for your contributions.

PR #384, opened 2015-04-30
Fix Animation on Safari

iaesaudi:
Added -webkit- before transform to fully support all webkit browsers specially Safari.


almasaeed2010:
I will add that to the LESS files since the changes you made will be overwritten by compiling the source code from LESS to CSS. Thanks for the suggestion!

PR #385, opened 2015-04-30
Fix Animation on Safari

iaesaudi:
Added -webkit- before transform to fully support all webkit browsers specially Safari.


almasaeed2010:
Thanks.

PR #391, opened 2015-05-01
Fixed text inside breadcrumbs

suvjunmd:


almasaeed2010:
And thanks again! :+1:

PR #404, opened 2015-05-09
changed q in jQuery

citrusfizz:
in case sensitive enviroments, the master branch is broken with this file name.


almasaeed2010:
Thanks.

PR #406, opened 2015-05-09
Changed text of version to 2.1

githubjeka:


almasaeed2010:
The dashboard version is not related to the version of the template. We may have Dashboard 3.0 in AdminLTE 2.2 for example. 

Thank you for the contribution. :+1:

PR #412, opened 2015-05-09
Removed redundant link to Ionicons

suvjunmd:
This will eliminate unnecessary downloads.


almasaeed2010:
I am keeping IonIcons in all pages since I may use them in navbar dropdowns in the future. 

Thank you for the contribution.

PR #426, opened 2015-05-14
Add an option to tweak the animation speed

hienning:
Sometimes we need to customize the animation speed. It will be convenient if there is such an option.

### DIFF
--- a/dist/js/app.js
+++ b/dist/js/app.js
@@ -35,6 +35,8 @@ $.AdminLTE = {};
  * Modify these options to suit your implementation
  */
 $.AdminLTE.options = {
+  // Animation speed for sliding
+  animationSpeed: 'fast',
   //Add slimscroll to navbar menus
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Thanks for the contribution. I will look into this soon. We may need to add this to animated elements, rather than only the sidebar if there are other elements that use animation.

YasserHassan:
What about making this option a part of the element HTML via a `data` attribute such as `data-aspeed` and the function looks for it on each animated element and uses if found. If not, a default or global setting is used instead.

almasaeed2010:
There aren't many elements that use JS animation so there isn't a need for that. But it is nice to have an option in AdminLTEOptions for the elements that do use animation.

almasaeed2010:
Options have been added to master.

Thank you for the contribution.

hienning:
Good work and appreciated 👍

PR #436, opened 2015-05-17
Updated list of skins with skin switch and removed whitespace from end of lines

freezy-sk:


almasaeed2010:
Hi,

Why did you change this line:
```html
<link href="../dist/css/skins/skin-blue.min.css" rel="stylesheet" type="text/css" />
```
To this line:
```html
<link href="../dist/css/skins/_all-skins.min.css" rel="stylesheet" type="text/css" />
```

You need only to choose one skin for your project and the blue skin is the one I chose for the documentation. There is no need to increase the load by getting all of the skin files.

sniper7kills:
@almasaeed2010 If you look at documentation/docs.js @freezy-sk  added 

    //Skin switcher
    $('#layout-skins-list a').click(function(e) {
        e.preventDefault();
        var skinName = $(this).siblings('code').text();
        $('body').attr('class', skinName + ' fixed');
    });

Which would change the skin theme requiring all of the skins not just the blue one. I think this may be something nice to include into the examples instead of or in addition to just the documentation.

freezy-sk:
@almasaeed2010 as @sniper7kills wrote, it's because of skin switch functionality which I've added to docs for better showcase of skin variations.

almasaeed2010:
Thanks!

Issue #425, opened 2015-05-13
Handling box pre/post remove events


YasserHassan:
@almasaeed2010 is there currently a way to detect pre/post box removal (by clicking the remove icon) to perform related actions such as storing that state in a session or connecting to the server or prevent the removal if the update failed, or let the user confirm before the action box removal. Handling other events is also useful such as updating an initially collapsed box via Ajax when expanded but I believe this could be handled via jQuery events.

almasaeed2010:
There isn't currently a way to do that through AdminLTE's app.js. However, I like the idea of adding before/afterRemove and Collapse handlers. I will have to look into it. Thanks for the suggestion.

pbihler:
I submitted a pull request to trigger events after the action is completed.

It's not trivial to privde a stable `before-...` handler, so it's easier to overwrite the standard AdminLTE handler:
```js
var realBoxRemove = $.AdminLTE.boxWidget.remove;
$.AdminLTE.boxWidget.remove = function(element) {
  var box = element.parents(".box").first();
  // decide if to really remove the box, then:
  realBoxRemove.apply(this,arguments); // forward call to original handler
};
```

RoelRoel:
This what I needed! Now an easy way to restore the states...

PR #449, opened 2015-05-21
Trigger jQuery-events upon box collapse/expand/remove.

pbihler:
Solves the issue #425.

In addition, it allows to register a handler that really removes "removed" boxes from the DOM (as explained in the documentation of the box events).


YasserHassan:
Thank you @pbihler.

almasaeed2010:
This would work just fine. However, I'd like to namespace the events to ```lte.box.expand```, ```lte.box.collapse```, and ```lte.box.remove```. Also, these are all post event operations. We'll also need pre event operations. Those would be named ```lte.box.expanded```, ```lte.box.collapsed```, and ```lte.box.removed```. 

Note: the pull request can't currently be merged since it conflicts with recent changes according to Github.

pbihler:
How do you plan to implement pre events? Simply triggering them before the slideUp call does not work, since they are delivered afterwards. 

One possibility would be to rely on a callback, but that is incompatible with the events concept.

Pascal

Karmalakas:
I'm in need of the events also, but one more thing would be really useful - trigger expand/collapse/remove.

Issue #417, opened 2015-05-10
Incorrenct Bower version


YasserHassan:
The version mentioned in bower.json file says 2.1.0. This seems to be repeating so I'm evaluating if it's better to use the PHP Composer package instead as it seems to be more reliable in this area.

YasserHassan:
@almasaeed2010 I am thinking of a way having some sort of automated version update from a single location that holds the current version as a string and triggered either automatically (on file update) or manually to update that version whenever required (e.g. HTML template, bower.json file) alone with anything else that may require repetitive manual update. This will make it consistent everywhere and also simplify the process. What do you think?

almasaeed2010:
This is very annoying specially when releasing quick-fixes or patches. I seem to always miss to update the version in some file.

YasserHassan:
Of course it'd be confusing to try to remember and keep track of those files. I'm thinking of two approaches to handle that. The first is a manual one by keeping a release checklist like:

Release List:
- Update version bower.json
- Update version in package.json
- Update version in index2.html
- ...etc

The other approach is using templates for those specific files. The checklist would be a very simple yet effective approach.

almasaeed2010:
That's a good idea. Thank you for providing this. I will create a similar list to help me remember before I release next versions.

YasserHassan:
@almasaeed2010 I am glad I could help.

Issue #289, opened 2015-03-13
Add a version with sass


stevenjlho:
I like sass than less. I hope that this project have a version with sass. Can to add this task to your todo list?

almasaeed2010:
Hello,
Unfortunately I am not very familiar with the sass syntax. It would also make it harder for me to maintain the template when it comes to updates. I apologize for the inconvenience.

stevenjlho:
I’m sorry to hear that. Anyway, thanks for your reply.

PR #454, opened 2015-05-23
SASS Support

karthilxg:
1. Added SASS support for AdminLTE styles.
2. Included Bootstrap-SASS version.
3. Updated Grunt task to run the SASS compiler.
4. Tested the Application with the new style.

Note: There are minor difference in the final output (AdminLTE.css) formatting, SASS updates the colors from #ffffff to #FFF etc.

https://www.diffchecker.com/mshpedmn


YasserHassan:
This is a good effort @karthilxg but I am concerned about continuity. My understanding is that if this commit is merged, both the LESS and SASS editions will need to be updated and tested separately from now on with each and every change to keep them both in sync. In #289, @almasaeed2010 mentioned that adding SASS will make it harder for him to maintain the template when updated. In #417 and other similar issues, it was obvious that even maintaining the version number among different package files like `bower.json` is hard for him to manage. With that in mind, I believe that adding SASS (which serves the same purpose of the existing LESS) will be counter productive and its time is better spent enhancing the template and adding new features. Probably a better approach is keeping the SASS as a separate repository.

karthilxg:
@YasserHassan Thanks for the response, since lot of people are moving towards SASS from LESS, I thought that having SASS support would add more popularity and contributions to the project, but I like the idea of maintaining the SASS version as a separate repository.

almasaeed2010:
@YasserHassan has a point. I noticed the interest in SASS increasing and I would love to have the template support it. To your knowledge, are you aware of any script that could automate the process? I know that Bootstrap developers support both SASS and LESS and it would be dull to code everything twice. Do you think they are using some kind of script that would compile SASS to LESS or vice versa?

Thanks @karthilxg for the contribution.

karthilxg:
@almasaeed2010 I have used [less2sass](https://github.com/ekryski/less2sass) script for conversion, as mentioned in the caveats section it's not 100% reliable, we might need to fix the SASS compilation issues manually. 

During the conversion most of the issues were occurred on the following places,
* extend tags
* Using !import overrides
* @inlcude tags (while invoking mixins the args are separated with semicolon)
* @import (reference) tags are not supported (https://github.com/sass/sass/issues/1317)

almasaeed2010:
This seems like it would require time to maintain and unfortunately I don't have that luxury. I appreciate your contribution greatly. Perhaps in a future release I'd make SASS the main preprocessor for AdminLTE.

karthilxg:
`Bootstrap SASS` repo maintains a `convert rake task` for pulling the changes from `bootstrap LESS` master branch and converting LESS files in to equivalent SCSS files.

Can the same process be followed for our case ? 

https://github.com/twbs/bootstrap-sass#upstream-converter

>Keeping bootstrap-sass in sync with upstream changes from Bootstrap used to be an error prone and time consuming manual process. With Bootstrap 3 we have introduced a converter that automates this.

>Note: if you're just looking to use Bootstrap 3, see the installation section above.

>Upstream changes to the Bootstrap project can now be pulled in using the convert rake task.

>Here's an example run that would pull down the master branch from the main twbs/bootstrap repo:

>rake convert
>This will convert the latest LESS to Sass and update to the latest JS. To convert a specific branch or version, pass the branch name or the commit hash as the first task argument:

>rake convert[e8a1df5f060bf7e6631554648e0abde150aedbe4]
>The latest converter script is located here and does the following:

>Converts upstream bootstrap LESS files to its matching SCSS file.
>Copies all upstream JavaScript into assets/javascripts/bootstrap, a Sprockets manifest at assets/javascripts/bootstrap-sprockets.js, and a concatenation at assets/javascripts/bootstrap.js.
Copies all upstream font files into assets/fonts/bootstrap.
Sets Bootstrap::BOOTSTRAP_SHA in version.rb to the branch sha.
This converter fully converts original LESS to SCSS. Conversion is automatic but requires instructions for certain transformations (see converter output). Please submit GitHub issues tagged with conversion.

almasaeed2010:
This is what I was looking for. I am not sure though if the converter supports everything that LESS uses. I will need to test it first. Thanks!

aguegu:
@almasaeed2010 

I am working on the scss version. And it is almost done. 

https://github.com/aguegu/AdminLTE

* all less have been convert to scss

And I can not help to add more trends/features in it

* bulid tool convert to gulp
* most plugins have been moved to bower.json
* the file structure have a big adjustment

Please check it out. And let me know what you think.

cesarizu:
That's cool @aguegu !. I've been using http://github.com/karthilxg/AdminLTE for SASS support and it's working without problems too. The only thing that I think it needs (as well as yours) is to declare variables with `!default` so that they can be overridden when needed (for customizations).

aguegu:
This could be a feature to work on. 

As this in a scss implementation, you may just rewrite the _varaibles in your favor and import the rest scss files. I am ok with refactorying the code, but I am so not a color tweaker... Changes only the syntax make it still trackable to the origin LESS styles. It is not that big different after all. I just completed my fork in a weekend. 

I wish @almasaeed2010 may look into this issue, there are too few `!default`, and too many `!important` in the source. 

I have register a bower package named admin-lte.scss, just stylesheet and js remained. All deps are left in config files. :) 

My fork has most completed. @almasaeed2010 , would you welcome my pull request? I sincerely doubt it, as there are too much changes. :(

And I have fix the js files by jshint. :)

PR #472, opened 2015-06-01
Comment metacode in index2.html

masteradhoc:
I commented the following metacode:
<meta content='width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no' name='viewport'>

""

### DIFF
--- a/index2.html
+++ b/index2.html
@@ -3,6 +3,7 @@
   <head>
     <meta charset="UTF-8">
     <title>AdminLTE 2 | Dashboard</title>
+    <!-- This tells the Browser, that your Website adapts to your width -->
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Hello,

Why did you choose to add the comment only in those 3 files?

Your contribution is greatly appreciated.

masteradhoc:
If you need them in some more files i'll do it for them too, no problem!
I think those three files are the startpoints for every developer who downloads your template. So i guess this line should be commented at least there. Should i do it for the other ones as well?

almasaeed2010:
Thank you for that. But you have a couple grammatical errors. The comment should be similar to:
```html
<!-- Tell the browser to be responsive to screen width -->
```

It will also be more appropriate if the comment was placed in all HTML pages except the documentation. If you find that a hassle, I can add the line at a later time.

Thanks!

PR #474, opened 2015-06-01
Comment metacode in starter.html

masteradhoc:
I commented the following metacode:
<meta content='width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no' name='viewport'>

""


masteradhoc:
This should clear some things up. See http://blog.javierusobiaga.com/stop-using-the-viewport-tag-until-you-know-ho

PR #475, opened 2015-06-01
Commented metatag

masteradhoc:
Hi @almasaeed2010 

Commented all pages, like discussed in #472 !

Thanks for your work
Brian


almasaeed2010:
Thanks!

PR #476, opened 2015-06-01
Typo in AdminLTE.css

masteradhoc:
Typo


almasaeed2010:
Thank you. We need to fix this in LESS too since it will be overridden when I recompile the LESS files.

masteradhoc:
Definatley! Will you do this?

almasaeed2010:
Done.

PR #489, opened 2015-06-05
Rtl Support

kossa:
Hello @almasaeed2010 
Thank you for your Admin LTE, it's very awesome, sure i'll use it for my projects ;)
I Already worked with SB Admin : https://github.com/IronSummitMedia/startbootstrap-sb-admin
And I added RTL support for this theme https://github.com/IronSummitMedia/startbootstrap-sb-admin/commits/gh-pages?author=kossa

And for AdminLTE, I'll use it for an Arabic project, so I must make it RTL, for that, yesterday I forked your repo, and I added RTL Support on new Branch called "rtl-dev", so you can test it easily ^^

I :
1. duplicated 'Blank page' into RTL Page
2. added Bootstrap RTL from cdn
3. added new file named rtl.css(and minifier version) 
4. Add link to new page on all old pages
5. That's All :)

Here is some screenshots :
Desktop 1 => http://i.imgur.com/d5aCiDU.png
Desktop 2 => http://i.imgur.com/GNDOfNv.png
Other Skin => http://i.imgur.com/fDJ9uWN.png
Ipad => http://i.imgur.com/9gbnwTT.png
Ipad2 => http://i.imgur.com/cftxi6K.png
Iphone => http://i.imgur.com/qS5cRXv.png
Iphone2 => http://i.imgur.com/975tRBg.png


masteradhoc:
Wow! This just looks great!
Think this calls for a new AdminLTE Version, after merging ;)

Thanks for this huge Contribution!

YasserHassan:
@kossa this is good work but I see you used CSS, not LESS as used in AdminLTE. This could be a major obstacle against merging your PR. Can you produce a LESS-based version instead to generate the RTL version without any manual enhancements or fixes?

almasaeed2010:
First of all thanks for the contribution!
I will test the code when I get a chance and let you know how it works on all supported browsers.

Have you used an automation script to convert to RTL? If so, does the script support the transform property? Thanks again.

almasaeed2010:
Regarding the LESS version, if an automation script was used then it is not really necessary since the development will be on the LTR main version and the script will convert the product to RTL. Thus, no need to redo everything in LESS files.

kossa:
@almasaeed2010 , No I did not used automation script to convert to RTL, but manually ;)
Also I did not used Less, bcz I think not need, it's like a hack , and I changed some properties not all Less file, Let me know what do you think :D

almasaeed2010:
Overall, it is a great work and probably the best RTL version of AdminLTE V2 until now. There are a few inconsistencies and development problems.

### Development Issues
Since AdminLTE grows continuously and without an automated LTR to RTL script, the work of development doubles in time. That is, I will need to check each component so that it supports both LTR and RTL. Unfortunately, I don't have the time. What you have changed did not address all the components and only fixed issues with the layout, which is great but, unfortunately, not enough.

### Inconsistencies 
- Border and position issue.

screen shot 2015-06-05 at 12 54 33 pm

- Border on the left.

screen shot 2015-06-05 at 1 00 09 pm

- Border radius haven't been flipped.

screen shot 2015-06-05 at 1 00 28 pm

- Padding and margining are incompatible with the main design.

screen shot 2015-06-05 at 1 00 40 pm

Those issues will be similar to ones that will exist on other custom AdminLTE components. Here is an example of using rtl.css with the default dashboard v2.

screen shot 2015-06-05 at 1 14 00 pm

In the table above, the headers are not aligned to the right. The products list have left-floating images and right-floating labels when it should be the opposite. In that sense, multiple other components will have issues that need correcting. 

### Preferred Solution
The best solution for supporting RTL would be to create or find an automation script for the task. This method allows us to create and maintain an RTL version with minimal effort. I previously tested several scripts for both LESS and CSS, the result was good but none of the scripts supported the ```transform``` property amongst a few other properties. Thus, the sidebars were not working properly. I am still looking for a good script, so if someone knows of one, please let me know. 

The other solution is with AdminLTE 3.0. Since it will use Bootstrap 4, many major changes will be made. I plan on creating and maintaining an RTL version manually if by that time we don't find a proper script. AdminLTE 3.0 may also use [PostCSS](https://github.com/postcss/postcss) or [SASS](http://sass-lang.com/) which will make creating an RTL version easier using conditional statements. Therefore, allowing us to keep both RTL and LTR code right next to each other.

As always, I am open to suggestions. If you think that you can continue developing the rtl.css to support all components and fix most issues, I am willing to give it a try and provide some help.

Thanks and keep up the great work!

kossa:
For Transform we can use it manually, if it's only the problem ;)

I found an awesome tool for sublime text : https://packagecontrol.io/packages/CSS%20RTL%20generator
I did not tried it yet :)

kossa:
I fixed the issues :
http://i.imgur.com/rU4Svnf.png

ahmadasm:
hi, i have followed your content but i have a problem in IE 10.
http://www.uplooder.net/

alnour-altegani:
hi I found command line tool named 'gulp-css-flip' depend on **node** and **npm** it can be used to **automate the process for rtl support** for adminlte . 

the user need to install it globally by this command after install node and npm
```bash
npm install -g gulp-css-flip
```
and then you need to rtl bootstrap first
```bash
gulp-css-flip path/to/bootstrap.css > path/to/dist/rtl-bootstrap.css
```
and then you need to rtl the admin-lte.css and skin.css
```bash
gulp-css-flip path/to/admin-lte.css > path/to/dist/rtl-adminlte.css
gulp-css-flip path/to/selected-skin.css > path/to/dist/rtl-selected-skin.css
```
the last step is to set dir property to rtl in the html layout
like
```html
<html dir="rtl">
```
and it is done

lxsameer:
#618

NassarX:
@alnour-altegani Thanks man , it works well now ;)

alnour-altegani:
@NassarX you welcome. did you tested the tooltips and popovers ?

PR #503, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "AdminLTE",
-  "version": "2.1.2",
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Did not know that. I thought since we can search their registry that it was built and the ```version``` property is used. Thanks for letting us know!

kkirsche:
Yep. Also just an FYI there are discussions to move bower to NPM anyway so I wouldn't really plan to support it for ever: bower/bower#1520

Issue #523, opened 2015-06-19
Box containing form fails to collapse


fschaeffer:
I followed your example and included a form into a box using the box-body for the fields and box-footer as container for the buttons.

I then tried to add a collapse button in the box-tools. The collapse fails to work, as in app.js only the direct descendant of box is searched for box-body or box-footer. 
Wrapping the whole box in form-tags also fails, as e.g. `ENTER` triggers the collapse instead of submitting the form.
Adding form in the ruleset for identifying the to-be-collapsed element in app.js solves the problem.

```html
<div class="box box-info">
    <div class="box-header with-border">
        <h3 class="box-title">Horizontal Form</h3>
    </div><!-- /.box-header -->
    <!-- form start -->
    <form class="form-horizontal">
        <div class="box-body">
            <div class="form-group">
                <label for="inputEmail3" class="col-sm-2 control-label">Email</label>
                <div class="col-sm-10">
                    <input class="form-control" id="inputEmail3" placeholder="Email" type="email">
                </div>
            </div>
        </div><!-- /.box-body -->
        <div class="box-footer">
            <button type="submit" class="btn btn-default">Cancel</button>
            <button type="submit" class="btn btn-info pull-right">Sign in</button>
        </div><!-- /.box-footer -->
    </form><!-- /form -->
</div>
```
and then in app.js 

```javascript
var box_content = box.find("> .box-body, > .box-footer, > form > .box-body, > form > .box-footer");
```

fschaeffer:
Fixed in #524

Issue #321, opened 2015-04-03
Embed boxes collapse


Ryxeuf:
When you have embed boxes that you can collapse, if you collapse a child, then collapse its parent, then expand the parent, th child is expanded too but with the fa-plus icon instead of fa-minus.

Expanding children could be an error. 
Instead of using .find() on line 442, using .children() fix that issue.

almasaeed2010:
Could you please post the html for the embedded boxes. I will work on it soon. Thanks!

Ryxeuf:
<code>
<div class="box box-success">
    <div class="box-header with-border">
        <h3 class="box-title">
                Scene 1
        </h3>
        <div class="box-tools pull-right">
            <button data-widget="collapse" class="btn btn-box-tool"><i class="fa fa-minus"></i></button>
        </div>
    </div>
    <div class="box-body" style="display: block;">
        
    <div class="box box-success collapsed-box">
    <div class="box-header with-border">
        <h3 class="box-title">
                Plan 1
        </h3>
        <div class="box-tools pull-right">
            <button data-widget="collapse" class="btn btn-box-tool"><i class="fa fa-plus"></i></button>
        </div>
    </div>
    <div class="box-body" style="display: none;">
            C'EST QUAND MEME plapla
    </div>
</div>
    <div class="box box-success collapsed-box">
    <div class="box-header with-border">
        <h3 class="box-title">
                Plan 2
        </h3>
        <div class="box-tools pull-right">
            <button data-widget="collapse" class="btn btn-box-tool"><i class="fa fa-plus"></i></button>
        </div>
    </div>
    <div class="box-body" style="display: none;">
            C'EST QUAND MEME plapla
    </div>
</div>

    </div>
</div>
</code>

Ryxeuf:
Don't know if it works for you to read the html

almasaeed2010:
You have to put your code within three copies this symbol ``` ` ```. Review this link please:
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code

almasaeed2010:
I edited it for now to display the code.

almasaeed2010:
Fixed in v2.1.0a

PR #524, opened 2015-06-19
Allow collapsible form in box

fschaeffer:
See issue #523

### DIFF
--- a/dist/js/app.js
+++ b/dist/js/app.js
@@ -555,7 +555,7 @@ function _init() {
       //Find the box parent
       var box = element.parents(".box").first();
       //Find the body and the footer
-      var box_content = box.find("> .box-body, > .box-footer");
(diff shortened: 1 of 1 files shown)


almasaeed2010:
What happens if we just remove ```>``` instead of adding the ```> form``` to the query?

fschaeffer:
Hello,

I just tested your suggestion and it works, but I have no idea whether this would cause any other sideeffects. Was there a special reason why you chose the direct children selector?

almasaeed2010:
What happens is that we would reproduce an issue that we fixed long ago #321. 

I'll merge this PR then since it sounds like the best solution. Thank!

fschaeffer:
OK, for the moment it seems like the best solution, experienced some lag in the slideUp-effect but this was not reproducable anymore ...

BTW: I also added the data-collapse to the header so the whole header is acting as trigger for the collapse (sort of accordeon) which works nicely but is a very special case.

Thanks for merging

Issue #520, opened 2015-06-19
Add info-box-sm for compact info boxes


fschaeffer:
I really like the info-box, but from time to time they are just too big, so I created a smaller version for the info-box which is suitable for two lines of content and takes only half the width and half the size for the icon. If this is of interest I could prepare a PR.

General question: for a PR should the less AND css file be changed or will you create a new css file out of the build folder? What about documentation?

```
/*
 * Component: Info Box Small
 * -------------------
 */
.info-box.info-box-sm {
    min-height: 45px;
    small {
        font-size: 12px;
    }
    margin-bottom: 8px;
    .info-box-icon {
        height: 45px;
        width: 45px;
        font-size: 24px;
        line-height: 45px;        
    }
    .info-box-content {
        margin-left: 45px;
    }
}
```

Usage:
```
<div class="info-box info-box-sm">
    <span class="info-box-icon bg-primary"><i class="fa fa-shopping-cart"></i></span>
    <div class="info-box-content">
        <span class="info-box-text">Cart</span>
    </div><!-- /.info-box-content -->
</div><!-- /.info-box -->
```

almasaeed2010:
Please provide a screenshot as to what this accomplishes.

For a PR, a LESS file would suffice. I usually recompile before pushing. It's be great if you provided snippets in the documentation as an example, but it is not necessary since I update the docs every now and then.

fschaeffer:
Hello,

see attached screenshot for the result (I added some more CSS as in this special case I only need one line of content)...
![small_info_box](https://cloud.githubusercontent.com/assets/1931017/8277496/b2956cec-18c3-11e5-9cec-248b30c0aeec.png)

PR #537, opened 2015-06-25
Add smaller version of info-box

fschaeffer:
Add a variant of the info-box which has only half the icon-size, see #520


rafael-renan-pacheco:
:+1: such request, must push.

Issue #521, opened 2015-06-19
Enable box-tools in footer of a box


fschaeffer:
The box-tools component can already be used in the footer but the CSS for proper positioning is missing.

I added  the following CSS rules:

```
.box-footer {
    position: relative;
}

.box-footer > .box-tools {
    position: absolute;
    right: 10px;
    top: 5px;
}
```

If interested I would prepare a PR.

PR #538, opened 2015-06-25
Enable box-tools in box-footer

fschaeffer:
the box-tools should also be usable and proper configured in box-footer, see #521


almasaeed2010:
Hi,

Instead of rewriting the code for the box-footer class, we should reuse the preexisting rules by combining the selectors.

fschaeffer:
Do you have an idea how to do it? I tried some adjustments but all led to some misalignement of either the icon or the dropdown... so I included the whole box-tools from header...

PR #543, opened 2015-06-27
Update Bootstrap-social.less

danielsum:
Hi Abdullah, 

I've seen that bootstrap-social.less is not up to date as Pinterest for instance was not in the list... May I suggest to use the import from bower package to keep up to date this little stylesheet or use a little script to update this vendor package like this one : 

````bash
#!/usr/bin/env bash

curl -o build/less/bootstrap-social.less https://raw.githubusercontent.com/lipis/bootstrap-social/gh-pages/bootstrap-social.less
````

PS: I've just discovered your package and make a little donation, great work !

### DIFF
--- a/build/less/bootstrap-social.less
+++ b/build/less/bootstrap-social.less
@@ -1,12 +1,11 @@
 /*
- * Plugin: Social Buttons
- * ----------------------
+ * Social Buttons for Bootstrap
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Thanks!

Issue #544, opened 2015-06-29
Single-Page App: Gulp & Bower support


poliveira89:
On `README.md`you have already explained how to install AdminLTE with `bower`.
PS: Already identified that you use Grunt and that the `bower.json` does not contain any dependency like Bootstrap, jQuery, etc... It's right that the dependencies are already bundled, if it so it's a little bit concerning regarding the fact that associated with bower are a "reuse" philosophy.

Putting that last part aside, if I install this theme with Bower which is the best way to integrate with my SPA in Angular/Gulp? Should I create HTML elements with `.wrapper` and other classes defined on AdminLTE on my "empty" `index.html`?

On a personal note: Appreciate a lot that you released this template as OSS,  great design! :+1: :smiley:

almasaeed2010:
I am not sure I understand your question. The template doesn't provide full Bower support yet. It will in version 3.0. All the dependencies will be added to the Bower.json file. 

Please clarify your question in order for me to be able to help. 

I appreciate the compliment. Thanks!

erayaydin:
:+1: 

For now, bootstrap, jquery and other things downloaded static. I think `bootstrap`, `jquery` and others should be in `bower.json` as **dependency**.

PR #548, opened 2015-07-01
Dependencies moved to Bower

adriandmitroca:
I removed all third party packages used in AdminLTE so now they are downloaded as package dependencies. Now you can use these dependencies throughout your project without annoying file duplications.

Related with #544.


almasaeed2010:
This will be done with AdminLTE 3.0 since it affects the file hierarchy of the current implementation. 

I appreciate your contribution. Thanks!

PR #555, opened 2015-07-02
Update Slimscroll to V.1.3.6

masteradhoc:
Update Slimscroll to V.1.3.6


almasaeed2010:
The version we have has been modified to support [MS Touch Events](https://msdn.microsoft.com/en-us/library/dn433244(v=vs.85).aspx) which is required by IE 11. The updated version does not have that. 

Thanks for the contribution.

PR #564, opened 2015-07-08
Fix bug in documentation

01-Scripts:
Class  collapsed-box was missing for Expandable box example.


almasaeed2010:
Editing documentation has to be using the Grunt include method. Otherwise it will be overridden when compiled. 

Thank you for letting me know. I'll add the fix soon.

01-Scripts:
I didn't know that. Thanks for letting me know.

almasaeed2010:
Done in master. Thanks again!

PR #571, opened 2015-07-11
Add word wrap to drop down messages

Prophet731:
If the text is longer than the drop down width then it will be cut off. This fixes that.

### DIFF
--- a/build/less/dropdown.less
+++ b/build/less/dropdown.less
@@ -99,7 +99,7 @@
 .navbar-nav > .notifications-menu {
   > .dropdown-menu > li .menu {
     // Links inside the menu
-    > li > a {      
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Hello,

This can be achieved by removing the following lines:
```css
overflow: hidden;
text-overflow: ellipsis;
```

I will do that soon. Thank you!

almasaeed2010:
Turns out that doing that would cause design problems. The overflown text doesn't get aligned properly. The only way to fix that is to wrap the text with a div and add left margining to it. Currently, this would break backwards compatibility. This issue may be fixed in v3.0. We appreciate the suggestion.

Thanks.

PR #608, opened 2015-07-30
remove multiple selection height property

emnsen:
Height property can be auto or removed. I prefer to remove. because there is no height property original file

select2multiple

select2multiple2


fschaeffer:
+1 good catch

emnsen:
@fschaeffer thanks :blush:

almasaeed2010:
Thanks!

PR #625, opened 2015-08-08
Bootstrap HTML-Comment: Remove Version

Dennis14e:
When upgrading Bootstrap HTML files no longer need to be updated.


almasaeed2010:
Updating that is a simple search and replace through the IDE. I think it is a good way to keep track of the bootstrap version. I appreciate the effort!

PR #628, opened 2015-08-12
Updated Bootstrap version comment to v3.3.5

Dennis14e:
_"Updating that is a simple search and replace through the IDE."_ - #625
:stuck_out_tongue_winking_eye:

### DIFF
--- a/documentation/build/index.html
+++ b/documentation/build/index.html
@@ -5,7 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <title>AdminLTE 2 | Documentation</title>
     <meta content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no" name="viewport">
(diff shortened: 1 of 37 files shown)


almasaeed2010:
hahahahahahaha. That is funny. I forgot to update all of the versions until after I created the release. I really need to create a list of things to do before I create a release.

YasserHassan:
A release list would be good @almasaeed2010 but I'm afraid it could get lengthy and cumbersome with the progress of AdminLTE. My opinion is that DRY needs to be applied heavily to AdminLTE to save a lot of time and effort that could be redirected to evolving the project and enhancing the release quality. That's why I suggest to start adopting templating at this point of the project. It may delay the next release a little but once applied, we can see a good progress rate afterwards. It does not need be a sudden change, could be done as a side activity by whoever wants to contribute and performed to a couple of pages at a time until the whole template is converted.

Issue #638, opened 2015-08-17
HTML linter for Bootstrap projects


Leftyx:
I always use the linter for Bootstrap in my twbs project.
This is the script to include in the body (before the closing tag of the body):

      <script type="text/javascript">
            (function(){var s=document.createElement("script");s.onload=function(){bootlint.showLintReportForCurrentDocument([],{problemFree:false});};s.src="https://maxcdn.bootstrapcdn.com/bootlint/latest/bootlint.min.js";document.body.appendChild(s)})();
      </script>

It seems there are a few problems with the validation.

I've noticed you don't use a container inside a navbar as suggested in the [documentation](http://getbootstrap.com/components/#navbar).

almasaeed2010:
Hello @Leftyx,

That's a great suggestion and we will for sure consider it! I wonder if there is a grunt task for it though.

The container provides the needed paddings and margins in case the layout uses a specific width. It is also used to provide rows with the negative margins they need to properly fit the page. In our case, we added the margins to the necessary wrappers so there is no need to use a container.

almasaeed2010:
Added

kressly:
Has this been fixed ?

almasaeed2010:
It's not really an issue. It's just guidelines

kressly:
Ok thanks and Almasaeed u are still the best

PR #648, opened 2015-08-22
Added Bootstrap linter

freezy-sk:
Added Bootstrap linter requested in #638

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -124,6 +124,7 @@ module.exports = function (grunt) {
       }
     },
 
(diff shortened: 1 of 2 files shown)


almasaeed2010:
Thank you @freezy-sk

almasaeed2010:
The number of errors are too many. Most of them are regarding ```row``` not preceded by a container, which we have discussed in #638. I will look further into it.

freezy-sk:
@almasaeed2010 it will be nice if you solve errors marked with E003 and E014

Issue #652, opened 2015-08-24
Bootlint issues


freezy-sk:
## Warnings

- [x] **W005** - Unable to locate jQuery, which is required for Bootstrap's JavaScript plugins to work; however, you might not be using Bootstrap's JavaScript
- [x] **W007** - Found one or more `<button>`s missing a `type` attribute.
- [x] **W008** - Tooltips and popovers within button groups should have their `container` set to 'body'. Found tooltips/popovers that might lack this setting.
- [ ] **W012** - `.container` or `.container-fluid` should be the first child inside of a `.navbar`

## Errors

- [ ] **E003** - Found one or more `.row`s that were not children of a grid column or descendants of a `.container` or `.container-fluid`
- [x] **E009** - Button and input sizing within `.input-group`s can cause issues. Instead, use input group sizing classes `.input-group-lg` or `.input-group-sm`
- [x] **E014** - Columns (`.col-*-*`) can only be children of `.row`s or `.form-group`s
- [x] **E029** - Since grid classes apply to devices with screen widths greater than or equal to the breakpoint sizes (unless overridden by grid classes targeting larger screens)...
- [x] **E033** - `.alert` with dismiss button must have class `.alert-dismissible`

rafael-renan-pacheco:
```E043  Button classes (`.btn`, `.btn-*`, `.navbar-btn`) cannot be used on `<a>`s within `.navbar-nav`s.```

The profile popup causes this.

@almasaeed2010

PR #654, opened 2015-08-24
Fixed Bootlint warning W005

freezy-sk:
Fixed warning W005 listed in #652


almasaeed2010:
This the printable invoice. It doesn't need any javascript. In fact, I should remove the app.js file.

Thanks!

freezy-sk:
@almasaeed2010 ok, just let me know when to mark that item as solved

PR #666, opened 2015-08-30
Added .container-fluid to .content elements

freezy-sk:
This small change can be done because `.content` is extending `.container-fluid`.
Also it reduces number of lint erros in #652 from 163 to 75

### DIFF
--- a/index.html
+++ b/index.html
@@ -470,7 +470,7 @@
         </section>
 
         <!-- Main content -->
-        <section class="content">
+        <section class="container-fluid content">
           <!-- Small boxes (Stat box) -->
           <div class="row">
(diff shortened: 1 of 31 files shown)


almasaeed2010:
Since ```.content``` works in place of ```.container-fluid```, I don't see the reason for adding the extra class. If it's just because of the Bootstrap guidelines, it is not critical to follow them since it doesn't break the layout in any way.

I appreciate the contribution and apologize for any inconvenience! :+1:

freezy-sk:
@almasaeed2010 actually it doesn't. when you don't use `.container` or `.container-fluid` you must explicitly define `.col-xs-12` on cols if you want them to maintain default bootstrap behavior.

PR #667, opened 2015-08-30
Refactored navigation user-body in user-menu to semantically better justified nav

freezy-sk:
Refactored row and col navigation in `.user-menu`'s `.user-body` to semantically better justified nav.
This small change also solves 33 erros in #652


almasaeed2010:
That's a great idea! Unfortunately, I can't merge yet since it affects backwards compatibility. I will merge this with AdminLTE 2.4.0 since that version will contain major changes anyway.

Thanks!

freezy-sk:
@almasaeed2010 Just look at `dropdown.less`, it's maintaining backward compatibility

PR #680, opened 2015-09-11
Fixed dependency

farhanwazir:
ansi-regex and maxmin modules also required to accomplish grunt tasks.


farhanwazir:
If you received pull request before from my account then kindly ignore that and accept this request. Thank you

almasaeed2010:
Hello @farhanwazir 

Could you please explain why do we need these new dependencies?

Thanks!

PR #681, opened 2015-09-12
Fix wrong path for static images

yrchen:
The plugin "dataTables" contains the wrong path for static images. This PR will fix the errors.


almasaeed2010:
Thanks!

PR #717, opened 2015-10-04
Fix animation bug with collapsed sidebar and treeview menu.

zviryatko:
Fix bug with treeview menu when sidebar is collapsed and you clicked to menu which already visible.

### DIFF
--- a/dist/js/app.js
+++ b/dist/js/app.js
@@ -394,7 +394,7 @@ function _init() {
       var checkElement = $this.next();
 
       //Check if the next element is a menu and is visible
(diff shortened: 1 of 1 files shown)


iaesaudi:
Could you describe more about this bug, Because I can't figure the difference :disappointed:

zviryatko:
1. Toggle left sidebar to icons view
2. Hover the mouse over any menu item that has submenu
3. Click on opened submenu header

iaesaudi:
Still no difference :dizzy_face:

zviryatko:
When you click on "UI Elements" the submenu is going to collapsing
example

iaesaudi:
Great, I got it.

PR #719, opened 2015-10-05
Fixes #688

### LINKED ISSUE #688: fastclick is not working (300ms delay)
Hi! First of all I want to thank @almasaeed2010 for the fantastic template and for sharing it with us. 

I ran into an issue where fastclick isn't loading with a 300ms delay as a result. I've logged the issue with FastClick, but it's probably an issue with AdminLTE: https://github.com/ftlabs/fastclick/issues/409#issuecomment-141074229

You can try it out by going to adminLTE preview on an iphone and navigate through the menu. You'll notice a 300ms delay when clicking around menu items.

Any idea what can be the issue here?

iamgoodbytes:
Loading the minified version of fastclick doesn't work. My guess is that the FastClick object gets renamed in the minification process, making the check for "typeOf FastClick" useless and causing FastClick to never execute. Loading the unminified version of fastclick removes the 300ms click delay across the board.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1213,7 +1213,7 @@
 <!-- Slimscroll -->
 <script src="plugins/slimScroll/jquery.slimscroll.min.js"></script>
 <!-- FastClick -->
-<script src="plugins/fastclick/fastclick.min.js"></script>
+<script src="plugins/fastclick/fastclick.js"></script>
 <!-- AdminLTE App -->
 <script src="dist/js/app.min.js"></script>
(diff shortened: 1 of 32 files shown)


almasaeed2010:
That's interesting. I am gonna test your code soon and get back to you. :+1:

iamgoodbytes:
The easiest way to test is to open adminLTE in the ios simulator (comes free with xcode on macos) or by running it on your device. 😎

Issue #710, opened 2015-09-27
AdminLTE 3.0 Plans


almasaeed2010:
Hello everyone,

As you may already know, Bootstrap v4 is coming soon. Thus, a major update for AdminLTE is due with the release of BS4. For the time being, we will only take suggestions for AdminLTE 3. In other words, v2 will get bug fixes and patches but won't get new features.

**Quick glance at our plans for AdminLTE 3:**
- Full utilization of Bower for package management
- SASS instead of LESS
- Customized Bootstrap via variables to significantly decrease the number of (override) lines of code in AdminLTE.css
-  Front-end template using the easily customizable color-scheme of AdminLTE to create a sense of consistency and broaden AdminLTE's domain
- Color variables of AdminLTE will be removed and Bootstrap 4 color vars will replace them
  - That is to say, instead of @light-blue for example, we will use $brand-primary
- The box component will be replaced with the new BS4 cards
- The main header will use a custom navbar to avoid current issues with responsiveness
- The treeview in the sidebar will be rewritten to provide advanced features and options such as treeview search and expand only when a specific element gets clicked.
- App.js will be removed and ECMA6 will be used for each plugin to produce AdminLTE.js
  - Treeview.js for example will be its own plugin and app.js will only supply a demo use case
  - This gives developers the ability to create their own custom app.js and use only the relevant plugins
- I still prefer Grunt over Gulp so I will keep using it (this is debatable)
- The layout may make use of the new flexbox functionality that BS4 provides. Thus, reducing the number of calls to `$(window).resize(...)`

The product of these changes should make AdminLTE highly customizable and convert it into a framework level tool rather than just a template. Suggestions and contributions are highly recommended as always. If you know of new plugins, tools or ideas, please feel free to post them here.

jlamer:
"This gives developers the ability to create their own custom app.js and use only the relevant plugins" - perfect

"That is to say, instead of @light-blue for example, we will use $brand-primary" - but it would be nice if you could provide some precompiled .css files with old colors. You would save a lot of time for users who dont have experience with SASS.

YasserHassan:
[...]

PR #728, opened 2015-10-09
add sass

Gerardo-junior:
add sass


almasaeed2010:
Hello @Gerardo-junior 

As mentioned in #710, we are not adding new features to AdminLTE 2. We really appreciate the contribution!

SASS will be used in AdminLTE 3.

Issue #328, opened 2015-04-08
Dynamically adding options to sidebar


gremerritt:
Hi - first off I'm fairly new to both JS and jQuery so if this is something obvious, I apologize!

I'm trying to dynamically add items to the sidebar. My default menu is something like:
```html
<!-- Sidebar Menu -->
<ul class="sidebar-menu">
  <li id="trip1" class="treeview">
    <a href="#"><span>First Trip</span> <i class="fa fa-angle-left pull-right"></i></a>
    <ul class="treeview-menu">
      <li><a href="#">Details</a></li>
      <li><a href="#">Map</a></li>
    </ul>
  </li>
            
  <li class="addtotree"><a href="#"><span>+ Add Another Trip</span></a></li>
</ul><!-- /.sidebar-menu -->
```

Then I added the following to the app.js:
```javascript
$('.addtotree').click(function() {
  var $newTrip = $('<li class="treeview"></li>');
  var $tripDetails = $('<a href="#"><span>New Trip</span><i class="fa fa-angle-left pull-right"></i></a>');
  $newTrip.append($tripDetails);
	
  var $newTripDropdown = $('<ul class="treeview-menu"></ul>');
  var $details = $('<li><a href="#">Details</a></li>');
  var $map = $('<li><a href="#">Map</a></li>');
	
  $newTripDropdown.append($details);
  $newTripDropdown.append($map);
  $newTrip.append($newTripDropdown);
	
  $('ul[class="sidebar-menu"]').insertAt(-2, $newTrip);
	
  $.AdminLTE.tree('.sidebar');
})

jQuery.fn.insertAt = function(index, element) {
  var lastIndex = this.children().size()
  if (index < 0) {
    index = Math.max(0, lastIndex + 1 + index)
  }
  this.append(element)
  if (index < lastIndex) {
    this.children().eq(index).before(this.children().last())
  }
  return this;
}
```

This *almost* works... the newly created menu item ("New Trip") functions as I'd expect, (click it and it expands, click it again and it collapses.) However, the original menu item ("First Trip") doesn't. When I click it, it opens but then immediately closes again. I'm assuming this is happening because of my call to `$.AdminLTE.tree('.sidebar')` - I thought this was needed in order to enable to menu to expand, but clearly it has undesired consequences.

![In action...](http://g.recordit.co/3n4ExXmTFo.gif)

Any help would be appreciated - if there's an easier way of doing this I'm open to suggestions!

[...]

PR #729, opened 2015-10-11
Sidebar autoclose on turbolinks fixed. close #718

### LINKED ISSUE #718: Events firing twice or more
I'm using Turbolinks jquery to with AdminLTE and after each request events handlers execute more than once. After some investigation I found out that you binds you `document` inside a `$(function())` block. This is not a good idea generally and should we should avoid such mistakes. For instance, this example below can be a common occurrence and should be avoided:

``` javascript
/* BAD: don't bind 'document' events while inside $()! */
$(function() {
  $(document).on('click', 'button', function() { ... })
});
```

You should be binding your events outside a `$(function())` block. This will ensure that your events will only ever be bound once.

``` javascript
/* Good: events are bound outside a $() wrapper. */
$(document).on('click', 'button', function() { ... })
```

lxsameer:
It's a very simple yet effective fix which prevent events on document fire more than once when using Turbolinks or similar tools.

Fixes #718, #328

### DIFF
--- a/dist/js/app.js
+++ b/dist/js/app.js
@@ -388,7 +388,7 @@ function _init() {
   $.AdminLTE.tree = function (menu) {
     var _this = this;
     var animationSpeed = $.AdminLTE.options.animationSpeed;
-    $(document).on('click', menu + ' li a', function (e) {
+    $(menu).on('click', 'li a', function (e) {
       //Get the clicked link and the next element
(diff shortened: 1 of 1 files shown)


lxsameer:
@almasaeed2010 Can you please release a bugfix version which contains this PR ?

PR #730, opened 2015-10-11
Remove sidebar click handler, before assigning one

xvonabur:
I would like to propose a small change in assigning a click handler for sidebar items. The current code doesn't work well with the Rails Turbolinks and jquery.turbolinks gem. It fires multiple times, like in #328 issue. Maybe I'm wrong here, but it seems the proposed code is one of the best practices for those kind of problem. 

I found a solution [here](http://www.gajotres.net/prevent-jquery-multiple-event-triggering/).

### DIFF
--- a/dist/js/app.js
+++ b/dist/js/app.js
@@ -388,7 +388,8 @@ function _init() {
   $.AdminLTE.tree = function (menu) {
     var _this = this;
     var animationSpeed = $.AdminLTE.options.animationSpeed;
-    $(document).on('click', menu + ' li a', function (e) {
+    $(document).off('click', menu + ' li a')
+               .on('click', menu + ' li a', function (e) {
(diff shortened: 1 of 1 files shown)


lxsameer:
it's related to #718 and #729

xvonabur:
Yep. But my solution is slightly different. I think $(document).on ... is working with static and dynamic DOM elements (again, see #328). But I'm not sure about the solution in #718. My Javascript kung-fu is still not the strongest one, so I'm open to discussion.

lxsameer:
While your solution is ok, I think setting an event handler on `document` is not a good idea when we have a selector to work with. setting the handler on and element is easier to do on document and won't bother turbolinks too

almasaeed2010:
Resolved in #729

PR #734, opened 2015-10-14
Add new rails gem supporting AdminLTE sources

cpg:
Add a link to the new gem that uses less and it's more maintainable


lxsameer:
There is another gem which supports sass version and both LTR and RTL `admin_lte-rails` in:
/Yellowen/admin_lte-rails

cpg:
You can add it to the docs too.

I did not test RTL with the gem I provided. If the base code supports it, so will my (RacketLogger, or RL) gem.

However, here are some of the differences:
* the RL gem does not include any RL plugin yet, the one you mention includes everything
* the RL gem is ready for Rails 4.x. The one you mention seems to be 3.x
* the RL gem is meant to be easier to maintain by using pristine sources
* the RL gem will allow people to include the plugins they need. For example, I find that iCheck breaks a number of Ajax/UJS things in my app related to checkboxes, hence I cannot use that Yellowen gem.
* the Yellowen gem is quite heavy in that it requires a lot of stuff like datetimepicker-rails, chart, etc. The RL one is meant to 
* the Yellowen gem requires execjs, which may or may not conflict with other users

That said, maybe you can add that Yellowen gem to the docs too. Let people choose.

PR #741, opened 2015-10-15
Do not use absolute position for .pull-right on sidebar

### LINKED ISSUE #740: Multiple label on one sidebar element
With this code:

```html
<li>
    <a href="pages/calendar.html">
        <i class="fa fa-calendar"></i> <span>Calendar</span>
        <small class="label pull-right bg-red">3</small>
        <small class="label pull-right bg-red">4</small>
    </a>
</li>
```

You will see only the `4` label.

Seems to be related with the `position: absolute` css rules.

Soullivaneuh:
Fixes #740.

### DIFF
--- a/build/less/sidebar-mini.less
+++ b/build/less/sidebar-mini.less
@@ -133,9 +133,6 @@
 .sidebar-menu li > a {
   position: relative;
   > .pull-right {
-    position: absolute;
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Will check this then get back to you. Thanks!

PR #790, opened 2015-11-11
Using velocity.js for animations

ashwin01:
Hi,

This is a patch which uses [velocity.js](http://julian.com/research/velocity/) library for the JQuery animations, which will have a performance boost in mobile and tablet browsers.

For starters, I've modified slideUp and slideDown functions in app.js to use velocity.js over JQuery. Would love to hear your opinion regarding this. May I suggest you to also use this plugin in version 3?

Kind regards.


almasaeed2010:
Hello,

Although this is interesting, we can't currently add new features to AdminLTE 2. We are working on 3.0 and will consider velocity.js. We appreciate the contribution!

Thanks!

ashwin01:
I understand. No problem, will look forward to the new version. Thanks!

PR #812, opened 2015-11-29
Error z-index css

condor-bird:
Pixels do not have to be specified for z-index.


almasaeed2010:
Thanks for pointing the issue out. We have to edit the LESS files rather than the compiled CSS so that it doesn't get overridden later on.

PR #815, opened 2015-11-30
add a codeigniter with adminlte project

chaegumi:


almasaeed2010:
Hello,

The documentation is built using Grunt includes. The changes should be made to the files within the build directory and then grunt includes must be ran to compile the results. Your current changes will be overridden if you don't use that approach.

I will add the resource to the docs soon.

Thanks!

PR #856, opened 2016-01-02
Changed hardcoded source and destination directories to variables

szegad:
Changed hardcoded source and destination directories (build & dist) to variables (srcdir & dstdir) to help integration with frameworks & CMSes.
 I'm using this Gruntfile when I'm integrating AdminLTE with Laravel.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,9 +4,11 @@ module.exports = function (grunt) {
   'use strict';
 
   grunt.initConfig({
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Hello,

We appreciate the suggestion. However, this will break the current setup of AdminLTE, so we can can't accept the changes. Sorry for the inconvenience.

Thanks!

PR #888, opened 2016-01-23
Added npm to readme

suvjunmd:
Surprisingly, the existence of an npm package is not mentioned in the readme.


almasaeed2010:
Thanks. We just added that.

PR #889, opened 2016-01-24
add tiles widgets support

limodou:
This widget comes from Metro-UI-CSS http:/…metroui.org.ua/. Changes are:

1. add animations.less, colors.less, images.less, reset.less, tiles.less, transform.less, utiles.less, 1-5.jpg
2. change variables.less, add new color names, add @borderRadius used in tiles
3. change AdminLTE.less, remove google font, because it's not locally, add imports for new less files
4. change app.js, add tiles, fitImage widgets and initialize revoke for metro widgets, also add menu item click event process
5. change doc for index.html, index2.html for new tiles entry, and add tiles.html page, but not changed all pages

Note: carousel of metro is not the same as bootstrap, so the effect is not available. And I think metro design of carousel is better.


almasaeed2010:
Hello,

Thanks for the great work! We appreciate your effort. However, those too many features that we can't provide support for. Therefore, we can't merge this for now.

Sorry for the inconvenience!

PR #904, opened 2016-02-01
Improve tree click event handler

0xMatt:
Click event is not currently firing under angularJS implmentation. This fixes it without any breaking changes to the current functionality.


almasaeed2010:
Thanks!

PR #909, opened 2016-02-04
Fixed #908

### LINKED ISSUE #908: Sidebar menu losing state after collapsed up and down
This is the initial state of sidebar menu:
image

Then I collapse up:

image

Then I collapse down:

image

The active state of first sub menu is lost.

I wonder why remove the state.

yanni4night:
Fixed sub menu state.


almasaeed2010:
Hello,

This doesn't fix the issue since it won't remove the styling of all closed menus. Thanks!

PR #974, opened 2016-03-18
Fix selector in header.less

enumag:
I'm not sure why the color inherit and background transparent are here but they are causing some problems for me. Mainly the fact that they also apply to buttons in the `.navbar-nav > .user-menu > .dropdown-menu` element. Is this the correct way to fix it?


almasaeed2010:
Old version of AdminLTE. Please upgrade to the latest version before submitting pull requests.

Thanks for the contribution and sorry for any inconvenience.

PR #976, opened 2016-03-20
Fixed W3C Markup Validation errors and warnings

suvjunmd:
###### Errors:
1. Bad value **javascript::;** for attribute **href** on element **a**: syntax error
2. An **img** element must have an **alt** attribute, except under certain conditions.
3. Bad value **""** for attribute **name** on element **input**: Must not be empty.

###### Warnings:
1. Element **nav** does not need a **role** attribute.

###### Source:
https://validator.w3.org/


almasaeed2010:
Thanks for the contribution! We appreciate it.

PR #997, opened 2016-03-29
Fix white background flickering on footer when content is short

christianesperar:


almasaeed2010:
Changing these rules will probably affect other parts of the template. Have you tested it with all layout options?

Thanks!

PR #999, opened 2016-03-30
replace all '2015' to '2016'

### LINKED ISSUE #998: 2015 to 2016
This issue is trivial.

There are many "2015".
It should be replaced to "2016".

yu1ro:
close #998


bluemanos:
image

Changing the year in a AdminLTE html code is ok, but idea of changing year along external library is stupid! Especially if you are changing release date.
After next update of the libraries the dates will change back to 2015 - if they do not change the year of course ;)

yu1ro:
It's true, @bluemanos. :stuck_out_tongue_closed_eyes:
Thank you for your find them.
I will fix it.

bluemanos:
I suggest to close the issue, rollback your changes. And do this again. To avoid storing commits history of a changes that has been rollback.

yu1ro:
Yes, it is the best solution!

PR #1015, opened 2016-04-19
Update Date Range Picker

Fl0ux:
Date Range Picker was one year old.


almasaeed2010:
Thanks for the contribution!

PR #1034, opened 2016-05-05
Enable Angular2 template using starter.html copied to starter_angular2.html

marvinvperez:
I think and hope this could be helpful and useful as I found it myself. Many people are looking for awesome bootstrap template such as this but at the same time they want to hook it up to angular 2 like me. So with this pull request, we will have a very simple integration baseline without changing any existing code. Instead, I create an html out of starter.html and call it starter_angular2.html.
And finally, add a folder called "app" to house the angular2 boot and main app.
Cheers,
marvin


almasaeed2010:
As I don't fully understand AngularJS and, therefore, will be unable to maintain the code in future releases, I simply cannot merge this to our implementation. However, we greatly appreciate your effort and contribution and hope that someone finds this useful!

Thanks!

PR #1043, opened 2016-05-10
Layout Boxed bug when resize page

EdisonHarada:
When you use layout-boxed much times it gives an error of jQuery, because the _fix function in $.AdminLTE.controlSidebar, delegate itself to the resize method of page when it's layout-boxed, so after some resizing it turn to like 2k calls of the same method.

### DIFF
--- a/dist/js/app.js
+++ b/dist/js/app.js
@@ -476,6 +476,11 @@ function _init() {
           _this._fixForContent(sidebar);
         }
       }
+      
+      // When resize the page format sidebar position and height
+      $(window).resize(function () {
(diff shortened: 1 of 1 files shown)


almasaeed2010:
I understand what you did here but unfortunately it still doesn't fix the issue. I tested your code and the boxed layout still jams after a few resizes. We are currently working on a fix for that.

Thanks!

EdisonHarada:
Why not? Here it works :T

almasaeed2010:
Issue resolved.

Thanks!

PR #1056, opened 2016-05-22
Update ChartJS to 1.1.1

timwhite:
Update for newer ChartJS that fixes some bugs


almasaeed2010:
Thanks for the contribution! We have actually updated the plugin locally before we say this request. Therefore the master branch now has the latest version of ChartJS. 

Thanks!

PR #1070, opened 2016-06-08
remove external font reference from default shipped CSS; update docs

mirabilos:
users are now expected to either add that line to their own CSS or
provide their own fonts, but on the upside, this makes it possible
to actually use AdminLTE as shipped in production environments

fixes https://github.com/almasaeed2010/AdminLTE/issues/744

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,9 @@ Change log
 
 **For the most recent change log, visit the [releases page](https://github.com/almasaeed2010/AdminLTE/releases).** We will add a detailed release notes to each new release. 
 
+**v2.3.…:**
+- Users now must explicitly include Google fonts themselves
+
 **v2.3.1:**
 - Fix sidebar issue #676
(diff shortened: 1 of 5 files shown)


almasaeed2010:
Hello @mirabilos,

As discussed before, we don't see this as a necessary change. We greatly appreciate your contribution and effort to enhance AdminLTE and we apologize for the inconvenience.

mirabilos:
Directly embedding Google fonts (and analoglously CSS or JS) from CDN is an EU-GDPR violation that is liable for both damages and penalty (up to 6 months of jail time).

This was now confirmed in a recent ruling: https://rewis.io/urteile/urteil/lhm-20-01-2022-3-o-1749320/

Back when we needed to use AdminLTE we had to [fork](https://github.com/tarent/AdminDCP) it instead. I feel slightly vindicated that my prediction of this being a legal issue is validated.

PR #1073, opened 2016-06-12
Modified app.min.js , for issue# 1080

ObaidAshraf:
Modified the code which was removing "active" class inside e.slideDown code due to which selected link gets un-highlighted when menu in sidebar gets toggled.


almasaeed2010:
Hello @ObaidAshraf,

We appreciate the contribution, but the code must be changed in both app.js and app.min.js. In fact it is more important for app.js to reflect the changes since the compression is automatic using grunt.

Thanks!

ObaidAshraf:
Thankyou .. So should I add the fix in app.js as well?

almasaeed2010:
That would be great so then we can move on to testing the code and approving it if fully works.

ObaidAshraf:
@almasaeed2010 .. Ok .. I'll add the fix in app.js as well ..

ObaidAshraf:
@almasaeed2010 Sorry .. I was quite busy ... Did you add the fix in app.js? .. Sorry again, i couldn't keep up :disappointed:

almasaeed2010:
I only closed this because I thought you'd create a new pull request with the changes in app.js. Currently, since the changes are in 1 line, it is difficult to identify what you did exactly. But, once you add the changes to app.js, we will be able to identify and test your modifications.

Thanks!

PR #1082, opened 2016-06-23
Feature/load grunt config

gitarno:
The proposal, for v3, is to separate the configurations of modules of GRUNTFILE tasks into small modules and dynamically load with load-grunt-config task in a more dynamic way of manage tasks.

With this change:
It is only necessary to add new modules / file in the GRUNT-TASKS folder with the job name specified, so LOAD-GRUNT-CONFIG task will search all files and set small settings for specific modules.

Advantage: Gruntfile.js file is optimized, and only serves to define and register specific tasks / customized for development.

More details: http://firstandthird.github.io/load-grunt-config/


almasaeed2010:
Hello @gitarno,

I like the idea of modulation but unfortunately I can't currently merge because of conflict. Do you think you could pull the changes on your side and fix the conflicts?

We greatly appreciate your contribution!

gitarno:
Hello @almasaeed2010 ! 

All right, i pulled the commits, was necessary: 
* Separate modules of tasks
* Update modules of tasks
* Compatibilize old merged request

almasaeed2010:
Thanks for the effort!

Issue #1080, opened 2016-06-21
<li> lost the "active" class after Collapsing parent menu


whahuzhihao:
In sidebar area, when you expand treeview-menu after collapsing it , the child <li> element will lose the active class.

ObaidAshraf:
Check and apply Pull request #1073 ..

cyrilchapon:
in addition, any sibling menu item (same level) also loose .active

https://github.com/almasaeed2010/AdminLTE/issues/1209

cyrilchapon:
- Pulling last release (v2.3.6) doesn't fix
- Neither does current master (1a89a2feabad28fa53fc1bc994788721e46f16a2)
- Only @maxolasersquad version [https://github.com/Captiveyes/AdminLTE/commit/2a22f7c72fc24accfcc3c66606b3d186114a6ec5] fixes

Leodime:
Actually removing this line:
**_parent.find('li.active').removeClass('active');_**
 doesn't fix it. It just set every parent element to **_active_**.

Doing this instead worked for me: 
**parent.find('.treeview.active').removeClass('active');**

I hope it works for you too.

Issue #1002, opened 2016-04-04
Keep class "active" after close Multilevel menu in Left Sidebar


Onfealive:
![before](https://cloud.githubusercontent.com/assets/18258131/14242572/523c546a-fa7b-11e5-9e08-f854baf4ca29.jpg)![working](https://cloud.githubusercontent.com/assets/18258131/14242571/523bf22c-fa7b-11e5-84c8-950d55506e03.jpg)![after](https://cloud.githubusercontent.com/assets/18258131/14242573/5247b06c-fa7b-11e5-9d6e-6e835225dd37.jpg)

Thanks!

AEK-BKF:
:+1:  we need it :)

wmarques:
If one of you have a fix for this, it's an annoying bug :(

almasaeed2010:
Due to some neglect in my part, many issues weren't addressed for some time. Some of these issues are no longer relevant since a solution might have been found. Therefore, we decided to close all issues that are over a month old. However, if the issue still persists and you'd like help, then by all means repost your issue and we will consider it again. In the mean time, we will close this post.

Thanks for your understanding and we apologize for any inconvenience <3.

PR #1084, opened 2016-06-27
Fixed loss of "active" class after collapsing menu

blz-ea:
Sidebar menu won't lose active class when you collapse it.
New option: **sidebarKeepMenuOpen**. If true menus inside left sidebar won't collapse.

#1080
#1002

### DIFF
--- a/dist/css/AdminLTE.css
+++ b/dist/css/AdminLTE.css
@@ -541,7 +541,8 @@ a:focus {
   margin-right: 10px;
   margin-top: 3px;
 }
-.sidebar-menu li.active > a > .fa-angle-left {
+.sidebar-menu > li.open > a > .fa-angle-left,
+.sidebar-menu .treeview-menu > li.open > a > .fa-angle-left {
   -webkit-transform: rotate(-90deg);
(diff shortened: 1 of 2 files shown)


almasaeed2010:
Hello @blz-ea,

I tested the code but the effect you suggested is not present. To make sure I understand correctly, please explain what the code should do.

We appreciate your contribution!

blz-ea:
Hello @almasaeed2010,
Code should keep "active" class even if I select/collapse menus.
As it was described in issue #1002 when I collapse "active" menu or open some other menus "active" class disappears.

 **sidebarKeepMenuOpen** if set to **true** keeps menus open

screenshot from 2016-06-29 00 11 14

almasaeed2010:
I see! The issue is that the feature wasn't implemented in the app.min.js file. I will merge the request and compress the file but I will change the option name to reflect the actual feature which is disabling the according in the sidebar.

Thanks for the contribution!

almasaeed2010:
@blz-ea did you test both options? When I set the option to false, the option doesn't reverse itself. Could you test that please?

PR #1001, opened 2016-04-01
Sidebar autoclose. Duplicate of #729

xvonabur:
I'm experiencing a bug with autoclosing sidebar again after last update (to 2.3.3). Please, apply fix from issue #729 again.

PR #1094, opened 2016-07-01
Remove sidebar click handler, before assigning one. Fix sidebar autoclose bug.

xvonabur:
I've closed #1001, because it has problems with dynamic elements. I'm pretty sure this one doesn't.


almasaeed2010:
I merged this for testing purposes. I will minify app.js and push the changes. If anyone finds an issue or feels like this solution didn't work please let me know.

@xvonabur Thanks for the effort and the contribution!

PR #1104, opened 2016-07-07
Follow tree menu parent url if provided

Soullivaneuh:
## Subject

On some cases, we might want to no use tree view collapsing but keep a link on the parent menu item.

This make sense to follow the link if a real URL is provided (e.g. not `#`)
## Concrete example

Take this menu extract:

image

The sub-menu only appear when you go on a specific domain dashboard (in this case, `chevallier.fr`).

But the "Domains" parent item is a link to the domain list.

ATM, I can't go back to this list because this link is now a collapsing action.

This PR solve the issue.

### DIFF
--- a/dist/js/app.js
+++ b/dist/js/app.js
@@ -394,6 +394,11 @@ function _init() {
         var $this = $(this);
         var checkElement = $this.next();
 
+        // If the link has an url, just follow it.
+        if ('#' !== $this.attr('href')) {
+          return;
+        }
+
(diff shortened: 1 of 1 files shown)


almasaeed2010:
This issue has been fixed in v2.4-alpha. Sincere apologies for the extreme delay!

Thanks!

Soullivaneuh:
@almasaeed2010 Well, since when shoyuld we put a fix on a new minor alpha release?

PR #1117, opened 2016-07-13
apply lte variables button color to bootstrap button 

yupeng957:
if i change variables color (such as @light-blue ) and the button bgColor works just fine 
but when i click the button , the button bgColor work won't work (it's the bootstrap button bgColor)

apply lte variables button color to bootstrap button on (active & focus) background color


lenamtl:
I'm having the same problem with the btn-success using skin-blue

Issue #1152, opened 2016-08-06
Fixed header for small menu and empty page puts footer below the scroll


marisveide:
Hi!

I have created empty page and removed all menu items but one.
When adding "fixed" class to be <body> as in documentation, the header gets fixed height.
But the page is still scrolling vertically, and the footer appears at the bottom, outside the page, so it makes the page scrollable even if it doesn't contain any elements.

It appears that the min-height of the page is the window size + the header size. So that's why it's always a bit bigger than it should be.

How to fix that?

Thanks!

almasaeed2010:
Closing due to inactivity

PR #1153, opened 2016-08-06
#1152: Fixed header for small menu and empty page puts footer below t…

marisveide:
…he scroll. https://github.com/almasaeed2010/AdminLTE/issues/1152


marisveide:
Now the height of the content wrapper is correct.

PR #1177, opened 2016-08-31
Added resize vertical for textareas

dudnikgg:
I added few lines, in forms.less, file that allow textareas resize only vertical.
 horizontal resizing can break the page... (also added bower_components folder to .gitignore)


almasaeed2010:
Hello,

Thanks for the contribution. This option, however, should be optional. It would be great if you add that as a class (called ```textarea-resize```, for example) so that it doesn't apply to every textarea.

Thanks!

PR #1186, opened 2016-09-12
sliders bug fixes

manhhailua:
- use `.bootstrapSlider()` instead of `.slider()` because `.slider()` does not work right with react.js
- add attribute:type to `.ionRangeSlider()`: related issue at [ion.rangeSlider repo](https://github.com/IonDen/ion.rangeSlider/issues/343)


almasaeed2010:
Thanks!

PR #1193, opened 2016-09-21
Don't remove active class on collpasing parent menu. Fixes #1080.

### LINKED ISSUE #1080: <li> lost the "active" class after Collapsing parent menu

maxolasersquad:


almasaeed2010:
Unfortunately, we have to remove active in the current implementation since it won't collapse otherwise. In the next version of AdminLTE, there will be two different classes to differentiate between active and open.

Thanks!

PR #1202, opened 2016-10-01
Right To Left

hamid80386:
Hello
i was made new files for show right to left  side menu in rtl language like persian, ordu and arabic.

i hope like this :-)


hamid80386:
Hello Alma, as you see this branch has no conflict with your base branch,
this branch made for people by right to left language have menu in right by all changes need only by two step.
thanks for your awesome panel

almasaeed2010:
Hello @hamid80386,

We greatly appreciate the effort you put into this and we realize the need for an RTL version by an overwhelming amount of our users. However, your implementation contains many issues at the moment. Here are a few:

1- The control sidebar still appears on the right when it should be on the left
screen shot 2017-01-08 at 11 47 55 am

2- The tasks navbar dropdown menu  gets sliced
screen shot 2017-01-08 at 11 48 27 am


3- Some text using sidebar mini disappears
screen shot 2017-01-08 at 11 47 55 am

The problem is that with our implementation of AdminLTE v2 at this time is that it makes it really difficult to create an RTL version. In AdminLTE v3, we are going to use SASS conditionals to make sure that every component in AdminLTE is implemented correctly in both RTL and LTR. We appreciate your patience and look forward to your future contributions.

Sincerely,
Abdullah

hamid80386:
tnx for review
i will solve the problems and re-commit ;-)

PR #1203, opened 2016-10-02
Update chart.js version to 2.3.0

ernestre:


almasaeed2010:
Hello @ernestre 

Unfortunately, this will break implementations of ChartJS for our users. This would be ok if were to release a major update, but we are not doing that with v2 since we are working on v3. We appreciate your contribution.

Thanks!

PR #1207, opened 2016-10-03
Update full-callendar to 3.0.1 version

ernestre:


almasaeed2010:
Hello,

This is a major update that might break our implementation of the calendar as well as others who use the template. This would be ok if we were to release a major update of AdminLTE, but since were aren't, I can't merge this at the moment. Thanks!

PR #1233, opened 2016-10-28
Use local fonts

stmswitcher:
Hello @almasaeed2010!

First of all I'd like to thank you for such a great template! I've been using it in some of my projects and I really love it.

However, my recent project is for a small local area network with a very poor internet connection. Every bit is important here (like in good old days with dial-ups and stuff). And it takes quite a while to cache fonts from google servers for every client. So, I've added the Source Sans Pro font to the template so it could be used without the request to google.

Please, consider merging my changes, they're providing a good performance increment for the average load time of every page that uses AdminLTE.

Best,
Denis Alexandrov.


herberthudson:
:+1:

almasaeed2010:
Hello @stmswitcher, 

We greatly appreciate your contribution. However, we feel that this is not entirely necessary at the moment. It would be great if we can make this optional to opt-in. Since Google recommends loading fonts from their servers due to caching and speed, it is currently a better option to keep pulling the font from their servers. Your example is a very specific use-case in which you needed this feature.

Thanks for contributing to our project!

PR #1264, opened 2016-12-02
Fixed a typo

Strato:
It works because it's an alias. But it might be removed without warning.

While I'm here, being able to deactivate the treeview on the sidebar would be super nice, or have it as an _opt-in_ using a CSS class like *sidebar-as-tree*.


almasaeed2010:
Thanks!

PR #1273, opened 2016-12-12
updated the image boxed-bg.jpg location to dist

teddywest32:
There is no any `img` directory  inside `build` directory which is causing an issue when we build using `webpack` and `less`. If there is another method to build could you suggest me and decline the pull request.

Signed-off-by: Pralhad Shrestha <pralhad.shrestha05@gmail.com>

### DIFF
--- a/build/less/variables.less
+++ b/build/less/variables.less
@@ -4,7 +4,7 @@
 //PATHS
 //--------------------------------------------------------
 
-@boxed-layout-bg-image-path: "../img/boxed-bg.jpg";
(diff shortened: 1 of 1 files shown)


almasaeed2010:
The compiled LESS files go to ```dist/css```. Therefore, the path should be relative to the css file not the less files. 

Thanks! We appreciate your contributions!

teddywest32:
@almasaeed2010 But it would be preferred if the developer wants to use the `less` instead of using the compiled `css`. Like importing it to their own less files and compiling using `gulp` , `Webpack` etc.

PR #1279, opened 2016-12-15
Fix the layout maker height calculation

MileNanevski:
The $.AdminLTE.layout.activate() call should be wrapped with $(document).ready to make sure that the DOM is ready for js manipulation.

Currently, it's possible that the main content element is not rendered yet and doesn't have a height, so it's height will be set to min-height and the footer will occupy the rest of the page.


almasaeed2010:
We already use a shorthand for $(document).ready(). Please read this article https://learn.jquery.com/using-jquery-core/document-ready/

The use of ```$(function(){});``` is shorthand for ```$(document).ready()```.

Thanks!

PR #1283, opened 2016-12-19
Update README.md

maxtor3569:


almasaeed2010:
The user should always fork the repository then modify it when contributing code. This allows for pull-requests and pushing back to github. That's why we added "YOUR_USERNAME" instead of the master repo.

Thanks!

PR #1286, opened 2016-12-21
Fixed jumping buttons with Bootstrap tooltip

marcusportmann:
This pull request fixes an issue where buttons that are part of a Bootstrap button group jump on mouseover when a Bootstrap tooltip is added to them.


almasaeed2010:
Thanks <3

PR #1294, opened 2016-12-27
Fixes #1235 text overflow while in collapsed sidebar-mini

### LINKED ISSUE #1235: sidebar collapsed hover windows auto width?
screenshot 1

so as you can see up there.. the windows that appear when i hover into the collapsed sidebar, i think it is have a fixed size and not an auto width that will match how many character in the list written... 

so how to fix this kind of issue?

XzAeRo:
Adds the `white-space: normal` property to avoid text overflow in the submenu items while in collapsed sidebar-mini state.

### DIFF
--- a/build/less/sidebar-mini.less
+++ b/build/less/sidebar-mini.less
@@ -59,6 +59,7 @@
               position: absolute;
               width: @sidebar-width - 50;
               left: 50px;
+              white-space: normal;
             }
 
             //position the header & treeview menus


almasaeed2010:
We encountered a problem previously that required us to remove white-space: normal. Since the sidebar transition to sidebar-mini breaks the text, everything looks misaligned during the transition. Therefore, the best solution to this is to add a transitioning class that forces  to not break and apply it during the transition only. Unfortunately, your current solution would fix one thing and break another at this time. I will keep this open since a better solution could be implemented based on your suggestion. 

Thanks for the contribution!

XzAeRo:
Aaah I see. Thanks for the clarification!

santhoshkorukonda:
@almasaeed2010 @XzAeRo How about adding a class something like this in _sidebar.less_ file.?

```css
.sidebar-menu ul li a {
    white-space: normal;
    word-wrap: break-word;
 }
```

No matter whether it is collapsed or normal sidebar it works perfect in all cases.
I have checked with very long text in collapsed and normal sidebar and it is working fine.

Checkout the following images.

img2

img1

If this breaks something else let me know **I can fix** them.

Lionqueen94:
It does break when the parent title is too long, it works perfectly in the normal sidebar, but in sidebar mini the wrapped part of the title is only half readable because the children cover it.

Sauerbrei:
For anyone who might face this issue as well, maybe this will work for you:

```
.treeview-menu {
  white-space: normal;
}

.treeview-menu span {
  display: inline-flex;
  word-break: break-word;
  width: 80%;
}
```

results in

image

guillecardozo:
One simple solution is to use a br tag on long texts.

>  <li class="nav-item">
     <a href="@Url.Action("ReservasTomadasParaUnPeriodo", "Reserva")"
        class="nav-link">
         <i class="nav-icon far fa-dot-circle nav-icon"></i>
         <p>Reservas tomadas <br> para un período</></p>
     </a>
 </li>

 
 Image attached as example.
Captura de pantalla 2025-07-01 131312

PR #1301, opened 2017-01-03
Update index.html

loekiedepo:
Yayra package hasn't been updated since the release of Laravel 5.0.

Jeroen Noten provides an extremely easy install, his downloads have skyrocketed over the past months and he responds almost instantly on issues and pull requests. Very good contributor and I think he should be named here.


almasaeed2010:
We don't have to remove the first package to add the second. Also, we need to add this to the documentation/build/include/implmentations.html since this is then compiled using Grunt to the index.html file. I'll add that soon. 

Thank you!

almasaeed2010:
Added in latest commit.

PR #1354, opened 2017-02-05
Update Timepicker

frdteknikelektro:
With the new one, the widget is not hidden in scrolling container


almasaeed2010:
Hello @frdteknikelektro,

Thanks for sharing this. Could you please specify from which version to which version is the update?

Thanks!

frdteknikelektro:
from, I don't know, maybe below v0.3.0 to v0.5.2

frdteknikelektro:
I think this will not harm

PR #1359, opened 2017-02-09
Added CoPilot to example implementations. 

misterGF:
This is a version of AdminLTE that makes it easy to use Vue.js. It has been catching good traction. Would be honored if it were included in the implementation section of the documentation.


almasaeed2010:
You have to add this line in `documentation/build/include/implementations.html`. Otherwise, it will be overridden by Grunt-include.

Thanks!

misterGF:
Fixed. Thanks.

PR #1376, opened 2017-02-23
Do not remove child active class when collapsing/expanding parent in treeview

kspearrin:
Steps to reproduce the bug:

1. Visit https://almsaeedstudio.com/themes/AdminLTE/index2.html
2. Click "UI Elements" to expand that parent menu.
3. Click the click menu option "Buttons" to highlight it as the `active` page (white).
4. Collapse the "UI Elements" parent menu.
5. Expand the "UI Elements" parent menu.
6. Observe that the current active page "Buttons" is no longer highlighted white with the `active` class.

This PR fixes this bug by not removing the `active` class from a child just because the parent was collapsed and reopened.


almasaeed2010:
This issue has been resolved in v2.4 (currently in alpha stages). Thanks for the effort and contribution.

PR #1385, opened 2017-03-01
Add an option to enable/disable sidebar to behave like accordion

faizfizy:
Current sidebar behave like accordion (current active li will collapse when another menu is clicked). This PR add an option to enable/disable the sidebar to behave like accordion.


faizfizy:
Also resolve weird behavior where animation missing when collapsing current active li on first load.

PR #1396, opened 2017-03-07
Add allow open multiplex tree view option

steelywing:
and clean flow


almasaeed2010:
This feature has been added in AdminLTE2.4-alpha. The option name in the Tree plugin is `accordion` and takes a boolean.

Thanks for the effort and contribution!

PR #1416, opened 2017-03-20
Fixed input focus in sidebar after resizing (fixed layout)

prusk:
Highly related to this issue: https://github.com/almasaeed2010/AdminLTE/issues/427

On mobile when opening sidebar and focusing on an input (search bar in this case) the virtual keyboard will pop-up which will trigger the window resize (testable here: https://almsaeedstudio.com/themes/AdminLTE/pages/layout/fixed.html):

`  $(window, ".wrapper").resize(function () {
    _this.fix();
    _this.fixSidebar();
  });`

Slimscroll uses jQuery wrap and unwrap to add container to the slimscrollable content, this in hand removes the DOM and re-adds it which will cause the element to lose focus. So the virtual keyboard will pop up and then instantly disappear, making it impossible to type anything into search input. 

This code will check if input was focused in the sidebar prior to calling destroy and then init on the now resized sidebar. This is not optimal as it will cause a double refire, but it works. Possibly a more permanent solution is to add slimscroll only to the ul element containing the menu instead of the entire sidebar.

### DIFF
--- a/dist/js/app.js
+++ b/dist/js/app.js
@@ -297,6 +297,7 @@ function _init() {
       //Enable slimscroll for fixed layout
       if ($.AdminLTE.options.sidebarSlimScroll) {
         if (typeof $.fn.slimScroll != 'undefined') {
+          var focusedElement = $("input:focus", $(".sidebar"));
           //Destroy if it exists
           $(".sidebar").slimScroll({destroy: true}).height("auto");
           //Add slimscroll
@@ -305,6 +306,10 @@ function _init() {
             color: "rgba(0,0,0,0.2)",
             size: "3px"
           });
+          //Refocus input if it exists after slimscroll init
+          if(focusedElement[0]) {
+            focusedElement.focus();
+          }
         }
       }
     }


almasaeed2010:
A fix for this issue was submitted previously and will be released in v2.4 (currently in alpha). Thanks for the effort and contribution!

Muhamann:
Hello,

i opened a new Thread refering to this Topic:
https://github.com/almasaeed2010/AdminLTE/issues/1518

The issue is not solved with this fix... please look at my description. Thank you.

Muhamann

prusk:
Muhamann, my PR was not accepted as a better solution is probably inbound. My proposed solution works by remembering the focused input and then refocusing it after slimscroll init.
You can see the code here: https://github.com/almasaeed2010/AdminLTE/pull/1416/files

Cheers

Muhamann:
Thanks prusk, your solution really works.
I gonna drink a cold beer on you today here in warm Germany ;)

Thanky

PR #1423, opened 2017-03-28
Update sidebar.less

CUCO-Ultimate:
Added max-height: 45px in order to prevent users image from getting stretched, when image is tall


almasaeed2010:
`height: auto` preserves the ratio of the image. Images should originally have square dimensions to fit properly in the sidebar.

Thanks!

Issue #1182, opened 2016-09-02
Page not filling full height 


chlebta:
Hello,
I got an issue when i don't have many content in my page then it's not filling the screen and a blank space in bottom is appearing.
<img width="1280" alt="screen shot 2016-09-02 at 16 48 56" src="https://cloud.githubusercontent.com/assets/1204608/18208597/609383a4-7130-11e6-8e19-66cda3184924.png">

dudnikgg:
Hi, do you have errors in your debugger (F12)?

chlebta:
@dudnikgg  no erros in my debugger

ashep:
+1, I have the same trouble

zburk:
Can you upload your source file to look at? I've tried to reproduce not having content on the page but it still fills the entire page for me. Perhaps you missed a closing HTML tag?

<img width="1043" alt="screen shot 2016-09-08 at 11 32 38 am" src="https://cloud.githubusercontent.com/assets/5782327/18355797/47ad89fc-75b8-11e6-9320-8b0d3a56ddef.png">

parhs:
I have the same issue. I think the problem is that javascript calculation isn't correct.

https://github.com/almasaeed2010/AdminLTE/blob/master/dist/js/app.js#L261

Here I think it should be 

`  if (window_height >= sidebar_height + neg) {`

wildcat010:
Hi,

I am using angular2 and I had the same problem.

I solved it by just loading the "app.js" in my AppComponent.

import { Component } from '@angular/core';

@Component({
    selector: 'my-app',
    templateUrl: 'app/app.component.html'
})
export class AppComponent {
    constructor()
    {
        console.log('preparing to load...')
        let node = document.createElement('script');
        node.src = "../scripts/app/app.js"; // you app.js script
        node.type = 'text/javascript';
        node.async = true;
        node.charset = 'utf-8';
        document.getElementsByTagName('head')[0].appendChild(node);
    }
}


Hope it helps.

csotomon:
Just add app.js to page

duleorlovic:
I can't see last sidebar link. It is easy to reproduce. Open [blank page](https://almsaeedstudio.com/themes/AdminLTE/pages/examples/blank.html), collapse **Examples** menu and manually remove content from `control-sidebar` (it is on right and populated from demo.js).
Than try to resize the window until you do not see last sidebar link.
[...]

PR #1445, opened 2017-04-12
Set content-wrapper height so we see last sidebar link #1182

duleorlovic:
This fixes issue #1182 when user can not see last sidebar icon...


REJack:
#1182 Is already fixed so I'll close this PR

PR #1480, opened 2017-05-04
New RTL

hamid80386:
Hello 
I've solved the previous problems and requested new pull... 
Now RTL verison work fine...


REJack:
Please don't recreate your PR after new commits, they come automatically to your PR if you push them to your branch.

hamid80386:
Thank you for your note, i just saw some new problems and solved those...

MostafaSh:
Hi @hamid80386 
Thanks for your great job. 

I have a question.
How should i use your changes to have adminlte Right to Left?
I don't know how to use your changes.

would you help me please?
thanks

hamid80386:
Hello @MostafaSh 
this is easy
you must see index-rtl1.html or index-rtl2.html for have a sample.

just add rtl in the wrapper  class
if you want to dont change all pull-left classes to pull-right add "rtl-inv" to that class too, but we don't recommend it!
`<div class="wrapper rtl rtl-inv">`

MostafaSh:
thanks a lot @hamid80386!
now it looks really amazing !

but there is a problem. when I used data table, the "lengthChange" doesn't set in the
 corner and came to middle of row. 

how could i fix it?

image

hamid80386:
Dear @MostafaSh 
In RTL we look at main component in AdminLTE

In other page, we don't have any act, but if you want to make RTL for these component, you have easy job!

That need only some seconds!

Best Regards

almasaeed2010:
Hello @hamid80386 

First of all, thanks for the great work! We really appreciate it.

Unfortunately, implementing an RTL version is not this simple. At this point, your changes don't affect all of AdminLTE's components. For example, the custom drop down menus in the main navbar are not RTL, the control sidebar still appears on the right and borders of the light sidebar skins are appearing in the wrong direction.

Also, this is implemented without using any of the conventions or the tools to help us maintain it such as LESS.

Again, we really appreciate it but we can't merge this at this point.

Thanks!

PR #1499, opened 2017-05-18
Fix docs link

tomRedox:


almasaeed2010:
Hello @tomRedox 
Thanks for the reminder. We are actually migrating the docs to an online versioned system. This link should now work. I forgot to publish the docs earlier.

Thanks!

PR #1525, opened 2017-06-05
fix control sidebar top padding

kspearrin:
To see the the problem:
1. Go to https://adminlte.io/themes/AdminLTE/index2.html
2. Open the control sidebar
3. Set browser width to exactly 768px

Top padding adjustment media query was off by 1px. Window width 768px (iPad) would show the additional padding, when it should only show at < 768px.

This PR fixes that.


almasaeed2010:
Thanks!

PR #1526, opened 2017-06-06
replaced boxed-bg images with static color

mojoaxel:
I'm importing the less files from the build directory directly in my application (using webpack).
This allows to only import the components I need in my application.

This works fine except for the `@boxed-layout-bg-image-path` variable because it uses a relative path that is not availably.

To fix this I replaced the dummy-background-image with a static background color.

* boxed-gb.jpg was only used as dummy background in the boxed layout.
* boxed-gb.png was not used at all (as far as I could see).


andrex873:
This is a good change to be included in one release, those images are not needed, and solve the error when we use webpack.

almasaeed2010:
Merged. Thanks!

PR #1544, opened 2017-06-16
Use css "text-overflow" to turncate long texts

tutkun:
...


almasaeed2010:
I am sorry I can't review changes with the number of commits showing here :(

What is the issue you are trying to solve?

PR #1569, opened 2017-07-06
fix 'footer is not at the bottom' issue

Th3M3:
see https://github.com/almasaeed2010/AdminLTE/issues/1547#issuecomment-311759441 for details


almasaeed2010:
Hello @Th3M3 

Sorry for the delay! These fixes are applied to the dist file rather than the build files. They will be automatically removed when building.

Also, I am not sure we need a setTimeout here since it cases performance issue.

Thanks!

Th3M3:
Have my changes now been applied or do I need to open a new PR for this?

I have cloned the master branch yesterday but the issue is not solved.
The footer does still not align to bottom if the menu's height is bigger then the Window's height and the content's height is not.
Then the resulting gap is exactly the height of the footer.

Th3M3:
@almasaeed2010

PR #1584, opened 2017-07-19
border colour changed for .navbar-custom-menu

tutkun:
added right menu border colours. but, i don't yet know "less variable" that have for the other colors.

That is image


almasaeed2010:
I am sure how this fixes the issue. Could you please elaborate?

Also, all the `.less` extensions should be removed since LESS doesn't require them and this is the convention we follow.

Thanks!

PR #1587, opened 2017-07-20
Fixes corrupt looking table headers.

ccoltman:
Should also remove 700 from the Google Font API reference


ccoltman:
Actually I think removing 700 from the Google Font API reference is enough, as it's not the fault of the template, it's the fault of the font itself. I'll close this.

PR #1605, opened 2017-08-09
Don't hardcode slimscroll options.

ADmad:


almasaeed2010:
We'll need a way to pass these options instead of just removing them. Maybe we can introduce slimscroll options to the layout plugin.

ADmad:
Yes introducing options for slimscroll would be the right way but until then removing these hard coded values at least allows one to set required styling through css.

almasaeed2010:
I should test how it would look like with the those options first. Will get back to you soon.

Thanks

almasaeed2010:
doesn't seem to have a major effect by default

ADmad:
To me slimscroll's [default](http://rocha.la/jQuery-slimScroll#highlighter_631965) scrollbar handle seems much better than then the tiny 3px currently set :)

ADmad:
`3px` is way to small and `rgba(0,0,0,0.2)` too light for usability.

PR #1619, opened 2017-08-22
Composer installer support

HWGehring:
Introduces support for composer installer (see [here](https://getcomposer.org/doc/faqs/how-do-i-install-a-package-to-a-custom-path-for-my-framework.md)). With this feature PHP developers will be able to specify where they want AdminLTE installed by adding:

```
"extra": {
  "installer-paths": {
    "my/template/path/": ["type:template"]
  }
}
```
to their composer.json


almasaeed2010:
Thanks!

PR #1685, opened 2017-10-12
Fix box inside box collapsing error (#1681)

### LINKED ISSUE #1681: BoxWidget collapse issue when box inside box
When I have a box inside a box, if I colapse the child box, the parent box colapses as well. 
I found this issue only after I upgraded from 2.3 to 2.4

 In adminlte.js, the BoxWidget toggle method is called twice after clicking the collapse button of the child box. Child box triggers the toogle method for itself and also for parent box.
If you click the parent box collapse button, toggle method is only called once, and only the parent box is toogled.

I can't figure out why. It seams a simple fix if you can find the source of the problem.

bmanifold:
Previously, with a .box inside a .box, any clicking on a collapse/expand button in either .box would cause both of them to collapse/expand.  

This pull request makes sure parent and child boxes do not interfere with each other.

This should fix #1681

### DIFF
--- a/build/js/BoxWidget.js
+++ b/build/js/BoxWidget.js
@@ -23,6 +23,7 @@
   var Selector = {
     data     : '.box',
     collapsed: '.collapsed-box',
+    header   : '.box-header',
     body     : '.box-body',
     footer   : '.box-footer',
     tools    : '.box-tools'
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Thanks!

PR #1693, opened 2017-10-16
Change JS comment for better parsing

defro:
Hello @almasaeed2010 
Thanks to share your incredible work and so nice admin theme.

I need to parse your html files for my PHP project but I have one problem when I met comment of this style: `increaseArea: '20%' // optional`
Because my parser remove whitespace so I got something like this: `increaseArea: '20%' // optional});});</script>`

With my changes submitted I will got something like that: `increaseArea: '20%' /* optional */});});</script>`

Regards,
Joël

### DIFF
--- a/pages/examples/login.html
+++ b/pages/examples/login.html
@@ -89,7 +89,7 @@
     $('input').iCheck({
       checkboxClass: 'icheckbox_square-blue',
       radioClass: 'iradio_square-blue',
-      increaseArea: '20%' // optional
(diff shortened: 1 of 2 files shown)


defro:
Many thanks @almasaeed2010 !

almasaeed2010:
No problem! 

For future reference though, the minifying script should remove all comments before minification. Just in case (and this is likely) we go back to using the `//` style, you need to make sure your minification script does that.

Have a good day!

PR #1707, opened 2017-10-26
removed dist

Dylan-DPC:
Hey

You shouldn't be committing your dist folder. Here are the reasons: 

1.  It is redundant since these are built files and don't have to be tracked.
2.  If there are any changes in the build files, all the users will end up having unnecessary commits
3. The files are bulky which will result in slow clone & pull, push times (if dist files are changed)
4. The user is expected to generate the files when he is using the package


almasaeed2010:
Hello,

The dist file helps people download a fully compiled version without having to run all the build scripts. 

Thanks!

concept-core:
@almasaeed2010 You might wanna push the dist files in a separate branch.
I get why you want to keep them but I do agree with @Dylan-DPC.

PR #1738, opened 2017-11-22
Fix Missing Chart.js File

iqmal852:
This PR will fix missing Chart.js on Index2.html. This missing file causes by typo which is written Chart.js where it should be chart.js


almasaeed2010:
Thanks!

PR #1756, opened 2017-12-05
change plugin name in documentation from Slim Scroll to jQuery slimScroll

Adagio-cantabile:
`slim scroll` (https://www.npmjs.com/package/slimscroll) is another library which is different from jQuery slim scroll, it may cause misunderstandings.


almasaeed2010:
Thanks @Adagio-cantabile. The documentation is built with grunt. The changes has to be made to plugins.html. 

I've made the changes and pushed.

Thanks!

PR #1811, opened 2018-02-08
Include Front-End or Landing page example please

minedbp:
Hi Guys

If possible include simple front-end or landing page please for the next version.

Thanks
P O L


jacobzlogar:
@minedbp perhaps this would be better suited for the "issues" tab

PR #1817, opened 2018-02-16
Fixed multilevel menu issue on opening/closing

ivansammartino:
On multilevel menu I noticed an issue that I tried to fix.
The problem:
1. on the left sidebar menu, click on "Multilevel", then on "Level One" submenu, then on "Level Two" submenu to open all them.
2. now close "Multilevel" and re-open it: submenus are open but the arrow on the right is wrong (it should be down instead of left pointing). More, when first clicking on "Level One" or "Level Two" links, the right arrow turns down, but nothing more happens: you have to click a second time to "re-align" all the correct behaviors.

The pull request fixes these strange behaviors


daniele-athome:
It is usually good practice to just remove unnecessary lines of code instead of commenting them. History exists just for that :-) #my2cents

almasaeed2010:
Thanks!

oyeaussie:
I did this instead

`    tree.find(Selector.open).removeClass(ClassName.open)
    parentLi.removeClass(ClassName.open)
    tree.slideUp(this.options.animationSpeed, function () {
      **tree.find(Selector.treeviewMenu).slideUp()**
      $(this.element).trigger(collapsedEvent)
    }.bind(this))
  }`

This way, when you close the parent, all children are also closed.

daniele-athome:
@oyeaussie please use 3 backticks for improved readability.

PR #1847, opened 2018-03-28
Fixed an issue where height becomes NaN when there is no main-header

akamaya:
When .main-header is generated dynamically, height becomes undefined on onload.
NaN is assigned to min-height of .content-wrapper by undefined calculation.
I fixed the problem.


almasaeed2010:
This needs to be compiled first. Please run `grunt` to compile the js.

Thanks!

PR #1859, opened 2018-04-16
return the initialized editor

DomienJ:
I kept getting a 'wysihtml5Editor is undefined' error when using this code:

var wysihtml5Editor = $('#something').data('wysihtml5').editor;
wysihtml5Editor.composer.commands.exec('insertHTML','Some text to insert');

Then I went digging in the code and found that this : "this.editor =  this.createEditor(toolbarOpts);" on line 14613 does not return anything.
So when I add a return at the end of the createEditor function, it fixes my problem.


almasaeed2010:
This issue should be reported and fixed on the WYSIHTML repo not here. 

Thanks!

PR #1871, opened 2018-05-09
Fix for datatables paging

HarshGundecha:
Updated datatables minified js to latest from example found on [https://datatables.net/examples/styling/bootstrap4](url)


X-Coder:
Thanks for your fix, it did fix the issue for me too. This should be merged back into v3-dev.

But you have a wrong description:
It was a update for the minified version of datatables (not jquery) which was not updated to 1.10.16 yet as the full version already was.

almasaeed2010:
Data tables is managed by npm. The current version listed in package.json is 1.10.16. Please see the v3-dev branch.

Thanks!

X-Coder:
The problem is, if you install adminlte using only composer (packagist.org) without npm, you will end up with outdated dependencies.

Maybe a better way would be to setup (maybe [suggested](https://getcomposer.org/doc/04-schema.md#suggest)) dependencies in composer.json too, there are some composer installers available too.

PR #1877, opened 2018-05-21
Fix for .border-radius() double declaration mixin

ivansammartino:
Hello,
I found a double declaration for `.border-radius()` mixin that PHPStorm was identifying as error: I replaced it with a single declaration accepting a variable number of arguments, 1 to N, for dealing both with unique and different border radius

Thanks :-)

EDIT: I'm very sorry, this is my second absolute pull request and I still don't know how to pull single requests :-( Seems that my previous PR was attached too


almasaeed2010:
Hi,

LESS allows us to override mixins. Your changes seem like they'd break some border radius configuration.

Thanks for the contribution!

PR #1898, opened 2018-06-10
Login and register page - icons outside inputs

heemza:
Fix bug in login and register pages.

screen shot 2018-06-10 at 11 41 54

screen shot 2018-06-10 at 11 41 46

PR #1900, opened 2018-06-14
Update register.html and login.html (icons outside page)

XIIQ:
login and register page- icons outside page

image

image


almasaeed2010:
Already fixed in #1898 

Thanks for the contribution @XIIQ

Issue #1893, opened 2018-06-04
same text-color as background for nav-item:hover dropdown


X-Coder:
### Issue type:
- [x] Bug report       

### Environment:
* AdminLTE Version: v3-dev 
* Operating System: W10 
* Browser (Version):  Firefox/Chrome 

### Description:
If you click on dropdown menu of the tab example and hover the main nav-item labeled "Dropdown", it has same text-color as background, and is therefore not readable any more.
https://adminlte.io/themes/dev/AdminLTE/pages/UI/general.html

almasaeed2010:
Thanks for reporting @X-Coder

REJack:
This issue is fixed with #2110, you can test it in v3-dev branch.

PR #1939, opened 2018-07-17
same text-color as background for nav-item:hover dropdown

hendimarcos:
feel free to do it your way
https://github.com/almasaeed2010/AdminLTE/issues/1893


hendimarcos:
#1893

almasaeed2010:
Thanks!

Issue #1944, opened 2018-07-18
[3.0 Alpha2] Layout does not consider footer height when setting min-height for content-wrapper


ericenns:
### Issue type:
- [ ] Feature request 
- [ x ] Bug report       
- [ ] Documentation   

### Environment:
* AdminLTE Version:  3.0 Alpha 2
* Operating System:  Linux
* Browser (Version):  (Latest)

### Description:
Layout does not consider height of the footer when setting min-height for content-wrapper.

* View "Blank Page" example @ https://adminlte.io/themes/dev/AdminLTE/pages/examples/blank.html
* Note that you have to scroll down to view the footer

REJack:
This issue is already fixed

PR #1945, opened 2018-07-18
[BUGFIX] #1944 Layout min-height consider footer-height [v3]

ericenns:
Fix Layout content height to consider footer-height when setting min-height


almasaeed2010:
Thanks!

Issue #1976, opened 2018-08-22
BoxRefresh errors fixes in 2.4.5 (and older)


OldaHorak:
### Issue type:
- [ ] Feature request 
- [x] Bug report       
- [ ] Documentation   

### Environment:
* AdminLTE Version:  2.4.5 (and older)
* Operating System:  Win/Linux (not important)
* Browser (Version):  FF (not important)

### Description:
The BoxRefresh plugin does not work as described in documentation:

- the overlay doesn't appear, removing does JS error
- the content replaces all .box-body classed elements, not only the child of refreshed box
- the refresh starts with click on any part of refreshed box, not only the "trigger" button

### Fixes for JS code in adminlte.js ver. 2.4.5
**line 57:**
```
--  this.$overlay = $(options.overlay);
++  this.$overlay = $(options.overlayTemplate);
```
`.overlay` doesn't exist in `Default`, `overlayTemplate` has to be used instead

**line 73:**
```
--  $(this.options.content).html(response);
++  $(this.element).find(this.options.content).html(response);
```
do not replace all `.box-body` elements, only that is the child of `this.element`

**line 83:**
```
--  $(this.element).on('click', Selector.trigger, function (event) {
++  $(this.element).on('click', this.options.trigger, function (event) {
```
`Selector.trigger` does not exist, the trigger is given in `options`

**line 94:**
```
--  $(this.element).remove(this.$overlay);
++  $(this.$overlay).remove();
```
Issue #1767 - not pulled yet

I hope it can help you! Good luck!
O.

PR #1981, opened 2018-09-01
fix for issue #1976

OldaHorak:
Fix for issue #1976 

-  **AdminLTE/build/js/BoxRefresh.js** fixed,
-  **AdminLTE/dist/js/adminlte.js** recreated by `grunt`, and
-  **AdminLTE/dist/js/adminlte.min.js** recreated by `grunt` too.


OldaHorak:
I think, the statement `$('#my-box-widget').boxRefresh('load')` is not valid in general and it is the documentation mistake. The parameter of the `boxRefresh` method has to be a JSON defined array of options, not the simple string, as can be seen by the evaluation of the parameter later in plugin code.

In addition, there is a copy of the usage info as it is written in the source of the plugin:
```
/* BoxRefresh()
 * =========
 * Adds AJAX content control to a box.
 *
 * @Usage: $('#my-box').boxRefresh(options)
 *         or add [data-widget="box-refresh"] to the box element
 *         Pass any option as data-option="value"
 */
```
The `options` parameter has to have this form:
```
{
    source         : '',
    params         : {},
    trigger        : '.refresh-btn',
    content        : '.box-body',
    loadInContent  : true,
    responseType   : '',
    overlayTemplate: '<div class="overlay"><div class="fa fa-refresh fa-spin"></div></div>',
    onLoadStart    : function () {
    },
    onLoadDone     : function (response) {
      return response;
    }
}
```
as it is used in `Default` variable.

almasaeed2010:
Awesome. Thanks!

PR #2006, opened 2018-10-06
Add pi-hole's dashboard to implementations

Ewpratten:
pi-hole's dashboard has been using AdminLTE for a long time, yet was not in this list. So I am adding it.


almasaeed2010:
Thanks!

PR #2012, opened 2018-10-21
added Symfony 4 compatible AdminLTE bundle

kevinpapst:
This PR adds a link to https://adminlte.io/docs/2.4/implementations for a Symfony 4 compatible AdminLTE bundle.

The currently linked version has many flaws / does not work correct with the latest Symfony framework (version 4).

I started working on that bundle, as I desperately needed a solution one for my own project. Thought I share it with the community, as more and more people are upgrading to the latest framework version and need a solution for their theme.


almasaeed2010:
Thanks!

PR #2017, opened 2018-10-29
add gitattributes

zacksleo:
Added export-ignore files by .gitattributes,  as there are too many files which we do not need when import the package.


almasaeed2010:
Thanks!

PR #2026, opened 2018-11-26
fix($login-and-register): input border breaks when focus and with validation class

meteorlxy:
# Screenshots

## Before

qq 20181126175551
qq 20181126175400
qq 20181126175408

## After

qq 20181126192558
qq 20181126175315
qq 20181126175332

# Details

- remove `box-shadow` of input `:focus`
- add border on `input-group-append` according to the status
- fix boostrap issue [#25110](https://github.com/twbs/bootstrap/issues/25110) temporarily by adding border-radius manually

### DIFF
--- a/build/scss/_login_and_register.scss
+++ b/build/scss/_login_and_register.scss
@@ -38,10 +38,40 @@
   .input-group {
     .form-control {
       border-right: none;
+
+      &:focus {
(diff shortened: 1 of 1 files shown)


almasaeed2010:
Beautiful! Thanks!

PR #2029, opened 2018-11-28
Fix CONTENT height calculation by subtracting also footer height

dbugit:
The CONTENT height was being calculated based only on the header height.
The subtraction of footers height was missing.
By adding it the content height calculation was fixed.


meteorlxy:
Didn't change the src file? `build/js`

dbugit:
No...
My mistake...
I will close this and submit the correct PR

PR #2037, opened 2018-12-12
Calculate min-height of Selector.CONTENT by additionally reducing height of the footer.

Talagozis:
In order the footer to be displayed, the min-height of the layout need to consider the height of the footer.


almasaeed2010:
Resolved in a previous PR.

Thanks!

PR #2039, opened 2018-12-17
Fix CONTENT height calculation by subtracting also footer height

dbugit:


almasaeed2010:
Resolved in a previous PR.

Thanks!

PR #2045, opened 2019-01-05
fix the scrollTop uneffect in some case

DHclly:
according this blog ,https://www.cnblogs.com/starof/p/5238654.html , 
I know how to  solve that `scrollTop=0` is uneffect

### DIFF
--- a/documentation/docs.js
+++ b/documentation/docs.js
@@ -47,7 +47,8 @@ $(function () {
   })
 
   $($slideToTop).click(function () {
-    $('body').animate({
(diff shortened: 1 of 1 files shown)


DHclly:
the documentation can not scroll Top , I fix it.

if document has DTD , document.body.scrollTop always is 0。in this case ，should use `document.documentElement` , otherwise use `document.body`。

REJack:
If the bug still exists, please recreate your PR for the master branch, the v2 branch is only a stable copy of the lastest release.

PR #2064, opened 2019-02-21
fix version number

yusufmalikul:


REJack:
This is already fix due new releases

Issue #2082, opened 2019-03-24
slimScroll not loading on v3 branch


jorgevilaca82:
### Issue type:
- [ ] Feature request 
- [x] Bug report       
- [ ] Documentation   

### Environment:
* AdminLTE Version:   3.0.0-beta 
* Operating System:  Windows 
* Browser (Version):  Chrome (Latest), Edge(latest), Firefox(latest)

### Description:
On load index.html the default OS scrollbar is shown instead of thin slimScroll

![scrool](https://user-images.githubusercontent.com/2745535/54882809-340c5480-4e34-11e9-85a5-6a1f25a79829.PNG)

PR #2084, opened 2019-03-24
slimscroll on V3 dev  #2082

jorgevilaca82:
slimscroll implementation o V3 branch


jorgevilaca82:
suggested implementation for bug mentioned on #2082 bug issue

PR #2091, opened 2019-04-06
Fix layout skin-black-light height inconsistence

yohaneslumentut:
If we switch to skin-black-light from the other skin, there will be a layout height changes (about 1px). By revert back to box-shadow the layout height will remain the same on every changing skin activity.


yohaneslumentut:
Thank you for your quick response.

puikinsh:
Thank you for your contribution!

PR #2178, opened 2019-07-18
Fix Select2 heights

maxwellmc:
I noticed that the Select2 elements were not styled to be the same height as the other inputs (they were noticeably shorter). This fixes the heights for consistency.

_Before:_
adminlte-select2-before

_After:_
adminlte-select2-after


REJack:
@maxwellmc Thanks for this fix, I also thought about adding [ttskch/select2-bootstrap4-theme](https://github.com/ttskch/select2-bootstrap4-theme) to AdminLTE v3 it supports also BS4's input group.

I'll add this fix for now, but I replace our own select2 style with ttskch/select2-bootstrap4-theme after releasing beta.2.

PR #2180, opened 2019-07-20
Changing module to work with the file system. Remove 'ncp'.

Leviasan:
My environment: win7-x64.
If deleted the folder 'plugins' and calling: ``` npm run-script plugins ``` I get an error
ENOENTmkdirError
Module 'ncp' cannot recursively create new folders for copying files from another directory into them.
Maybe this problem only windows system, but if using the module 'fs-extra' - this problem disappears.
Module 'ncp' is the oldest and has not been updated for a long time compared to 'fs-extra'.

### DIFF
--- a/build/npm/Publish.js
+++ b/build/npm/Publish.js
@@ -1,5 +1,5 @@
 const Plugins = require('./Plugins')
-const ncp     = require('ncp').ncp
+const fse     = require('fs-extra')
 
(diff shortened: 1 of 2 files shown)


REJack:
👍 that's awesome I doesn't had time to check why it breaks, it occured on macOS too.

PR #2259, opened 2019-09-07
Fix for table-valign-middle doesn't work

milkcocoa:
At v3-beta1, `.table-valign-middle` class worked correctly, but now it doesn't. I think this class is set at the `table` tag, so this fix is needed.


REJack:
Thanks for this PR

PR #2321, opened 2019-10-11
Update _select2.scss

dbugit:
The arrow down overrides the clear "x" button.
Removing this padding-right: 0 fix this issue.


REJack:
I see what you mean, thanks for this fix 😄

PR #2322, opened 2019-10-11
Update _select2.scss

dbugit:
By defining an heigh for multiple selection, the selection box doens't follow the contents heigh.
You can even test this on your demo page.
By removing this property this issue is fixed.


REJack:
I see what you mean, but removing height is not the best approach. I would change height to min-height and add a aditional fix to get the correct height like default inputs.

REJack:
I've added a fix in the v3-repo, you can check it out tomorrow with the new rc.5 release 😄

PR #2332, opened 2019-10-16
fixed content height calc with less sidebar content

REJack:


bobvandevijver:
Right, I was just looking into this! Is it possible to release a new RC with this change, so I can test it?

PR #2528, opened 2020-01-31
[Tables] Expandable table

bansalshashank:
**[Feature] Expandable Table**


REJack:
Sorry for the late response, I'll release soon v3.0.3 and after this I start with v3.1 and create a development branch then I check and merge this PR as first change in v3.1.

bansalshashank:
thank you

XhmikosR:
@REJack there are some HTML validation warnings introduced in this PR which are valid and should be fixed. Should I enable `Werror` so that the test fail for warnings too?

XhmikosR:
This is poor UI IMHO. There's at least no visible indication this is clickable. Accessibility-wise not sure how it can be even made proper.

It should have been thought better before merged IMHO. Also, no point in using single quotes in HTML attributes.

REJack:
>This is poor UI IMHO. There's at least no visible indication this is clickable. Accessibility-wise not sure how it can be even made proper.
> 
> It should have been thought better before merged IMHO. Also, no point in using single quotes in HTML attributes.

Feel free to create a Issue for it :) I'll look over it and update the style & html markup.

XhmikosR:
What I mean is that these kind of stuff should not land before being proper because then things get messy to maintain.

I thought about accessibility and the plugin(s) need to handle at least aria-expanded.

PR #2578, opened 2020-02-29
[ENHANCEMENT] Added color to dropdown header

c0fe:
Simple adjustment, added color to the dropdown header and made the font bold.

### DIFF
--- a/build/scss/_dropdown.scss
+++ b/build/scss/_dropdown.scss
@@ -113,6 +113,12 @@
   font-size: $font-size-sm;
   padding: .5rem $dropdown-item-padding-x;
   text-align: center;
+  color: #8a93a2;
+  white-space: nowrap;
+}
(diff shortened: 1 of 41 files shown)


REJack:
I doesn't merge this PR.

1. To change the Dropdown Header color you can simply add 
2. The `<strong>` isn't required (simply add `.text-weight-bold` to `.dropdown-header`)
3. HardCoded color codes aren't accepted and it's only one fixed color
4. The colors have no contrast (not good to see on bad monitors)

Also I personally like the white background and it's more like the AdminLTE v2 style.

For an css-free style change you could use simply this markup for the header
```html
<span class="dropdown-item dropdown-header bg-light">
  <span class="font-weight-bold text-muted">15 Notifications</span>
</span>
```

This will look like this
image

PR #2612, opened 2020-03-23
Overlay for tab-panels

andcarpi:
Don't know if i did it right, double check plz... 
There's a way to avoid the use of a overlay-wrapper? i'm not good in css....

image

### DIFF
--- a/build/scss/_miscellaneous.scss
+++ b/build/scss/_miscellaneous.scss
@@ -288,6 +288,26 @@
   }
 }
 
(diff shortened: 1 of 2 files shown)


REJack:
Nice add, could you add a demo inside the examples for this?

andcarpi:
Added a demo to the navbar sample page
overlay, overlay-dark and normal tab...

REJack:
Perfect 👍

PR #2630, opened 2020-04-02
Fixed height brand-link

dima-bzz:


REJack:
Did you tested the fixed brand-link height with the text-sm classes?

dima-bzz:
No

REJack:
I tested your changes, this will break by users the brand-link height when they use text-sm.
If you change this to a extra class I will merge this PR otherwise not.

dima-bzz:
If unchanged, the brand-link line does not match the navbar

brand-link

REJack:
Then you need to increase the negative margin around your image. or change this as a extra class instead of an overall override.

PR #2684, opened 2020-05-13
Default params demo

Oxicode:


REJack:
This will cause be every user a changed default style, that's why I won't merge it.

PR #2708, opened 2020-05-30
Cleanup Gemfile and add wdm on Windows

XhmikosR:
@REJack do you build the docs locally only? If so, maybe you don't need the `github-pages` gem which installs 85(!) gems.

But I'll leave that for you to tackle later. This patch removes redundant gems; `github-pages` gem includes those.

EDIT: also is generally good practice to include Gemfile.lock in the repo, like you do with package-lock.json. And perhaps specify a version for the `github-pages` gem; now it installs the latest all the time.

### DIFF
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -1,7 +1,4 @@
 source 'https://rubygems.org'
 
-gem "jekyll", "~> 3.8.5"
-
-gem "rouge"
-
-gem 'github-pages', group: :jekyll_plugins
+gem 'github-pages'
+gem 'wdm', '~> 0.1.1', :install_if => Gem.win_platform?


REJack:
Yeah I create the docs for the npm package locally and on the offical docs page seperatly 😄 I also considered to push the docs into the github pages before I've published them to the offical docs page. But right now I wont publish them to the github pages since the offical docs work nice.

I'll test if the build works flawless on the offical docs and then I merge it, but I guess it will work even without the github-pages gem.

XhmikosR:
IMHO you just need Jekyll, jekyll-seo-tag and wdm for Windows.

So, this would save 83 gems, which is huge. I can make a PR later after this one lands.

REJack:
👍 I'll test it later this evening (if I find good internet spots on the campsite 😃)

REJack:
Tested and works the same way as before 😄

Issue #2605, opened 2020-03-22
bootstrap-switch v3.4.0 have a bug


snake19870227:
<img width="1191" alt="截屏2020-03-2222 45 12" src="https://user-images.githubusercontent.com/49919227/77252490-13691b80-6c8f-11ea-9195-a1e86122e5ad.png">
1 使用了一个未知的变量，2 一定进入不了 if 块，导致初始化的时候设置 state 属性不起作用

snake19870227:
<img width="819" alt="截屏2020-03-2222 56 57" src="https://user-images.githubusercontent.com/49919227/77252691-858e3000-6c90-11ea-8e82-9e70802c2828.png">
上图为bootstrap-switch 3.3.4版本相同逻辑的代码，使用了options中的state。
https://github.com/Bttstrp/bootstrap-switch/releases bootstrap-switch并未正式发布3.4.0版本

REJack:
Please write issues in english, otherwise I can't help.

snake19870227:
<img width="1191" alt="截屏2020-03-2222 45 12" src="https://user-images.githubusercontent.com/49919227/77254067-b9218800-6c99-11ea-80ba-1cb3972d8144.png">
bootstrap-switch v3.4.0 have a bug.
1. An unknown variable is used
2. never enter the if block
Causes the initialized 'state' parameter to fail
<img width="819" alt="截屏2020-03-2222 56 57" src="https://user-images.githubusercontent.com/49919227/77254128-1c131f00-6c9a-11ea-8b15-8d0afa8151ec.png">
bootstrap-switch v3.3.4 Code with the same functionality

snake19870227:
It is recommended to use bootstrap-switch v3.3.4

snake19870227:
Sorry! I'm not good at English

REJack:
I've downgraded the boostrap-switch to 3.3.4 only even the 3.3.5 has the same bug.

PR #2710, opened 2020-05-30
Update bootstrap-switch to v3.4.0.

XhmikosR:
The previous version didn't support Bootstrap 4.x.


XhmikosR:
Now I see https://github.com/ColorlibHQ/AdminLTE/commit/533a76538c98c06be7328176dffa562720edf4ac#diff-b9cfc7f2cdf78a7f4b91a753d10865a2 but it doesn't mention why it was reverted?

REJack:
One of our users found a massive bug in v3.4.0 also it's only published on npm but there is no github tag for v3.4.0, reference issue is #2605.

XhmikosR:
I agree about it not being published on GitHub, but that could probably be something they missed. That being said, 3.3.4 isn't compatible with Bootstrap 4.x.

Has the issue been reported upstream?

XhmikosR:
It seems that project is poorly maintained. I'll close this PR, but better look for an alternative if the issues aren't fixed soon :/

https://github.com/Bttstrp/bootstrap-switch/issues/717

REJack:
Yeah I know it's not maintained right now from the author, also since I've decided to drop jquery with v4 (like BS5 will do), I'll need to search good alternatives for the most 3rd-plugins.

PR #2712, opened 2020-05-30
Update CI config

XhmikosR:
* remove the non-LTS Node.js 13
* specify `CI: true` environment variable
* switch to `npm ci` which checks for package-lock.json and it's intended for CI

I'm marking it as draft to check any failures and fix them. Also, ideally the other deps PRs should land first, so I'll wait to rebase this.

EDIT: the failures should be fixed after the deps PRs are merged, so let's wait a bit :) Also, CI should be faster after the deps PRs and this lands.

### DIFF
--- a/.github/workflows/node-install.yml
+++ b/.github/workflows/node-install.yml
@@ -3,11 +3,14 @@
 
 name: Node.js Install Test
 
-on: 
+on:
   pull_request:
   release:
     types:
       - created
+
+env:
+  CI: true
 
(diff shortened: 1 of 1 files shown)


REJack:
The workflow checks are blazing fast now 😃 that's really amazing

XhmikosR:
`npm ci` was introduced for this specific reason :) Also, always try to keep your devDependencies up to date :)

PR #2716, opened 2020-05-30
Doc gems

XhmikosR:
Requires #2708.

~~I went with the gem versions that were used before. For example Jekyll has a 4.x version, but that's for you to tackle later :)~~ I updated all gems to the latest versions.

PS. CI could run this stuff.
PPS. Better not squash the patches.

https://github.com/ColorlibHQ/AdminLTE/pull/2716/files?w=1


XhmikosR:
I updated to Jekyll 4.x too. But I notice some pages are broken locally, which seems to happen with current master too. Maybe you are missing a `baseurl` in `_config.yml`?

REJack:
> I updated to Jekyll 4.x too. But I notice some pages are broken locally, which seems to happen with current master too. Maybe you are missing a `baseurl` in `_config.yml`?

Yeah I noticed the same, I am considering removing the docs from the npm package. Then we can set a baseurl to the `_config.yml`, actually I set a baseurl in the build command for the online docs on ever build.

XhmikosR:
> Yeah I noticed the same, I am considering removing the docs from the npm package. Then we can set a baseurl to the `_config.yml`, actually I set a baseurl in the build command for the online docs on ever build.

We could work around the issues by using URLs relative to the server root, i.e. `/foo.bar` and get rid of `baseurl` logic. Either way this issue is unrelated to this PR :)

PR #2721, opened 2020-05-30
Simplify CI

XhmikosR:
Non-whitespace diff: https://github.com/ColorlibHQ/AdminLTE/pull/2721/files?w=1

EDIT:

I didn't add caching because it's super slow on Windows: https://github.com/actions/cache/issues/336

Might want to revisit later.


XhmikosR:
I didn't add caching because it's super slow on Windows: https://github.com/actions/cache/issues/336

Might want to revisit later.

Issue #2725, opened 2020-05-30
[BUG] fix SCSS variable files


XhmikosR:
It seems there's a file `_bootstrap-variables.scss` that has some custom overrides. This file should be removed and `~bootstrap/scss/variables` should be included instead.

Every variable should be in `scss/_variables.scss` without the `!default` flag, and this file, `variables`, should be included in the rest of the files. This is the proper way to do it, reducing duplication. :)

REJack:
I know, I didn't like this either but I don't know what the original author has changed in the variables. We can change this in the v3.1 development to publish soon a new version (v3.0.6) and start with the v3.1 development 😄

danny007in:
Any new idea for AdminLTE v4

craph:
does this bug even exist in v4 ?

puikinsh:
Closing as part of issue triage for v4.0.0-rc7. This issue predates the current v4 architecture and is either resolved, outdated, or outside scope. Please open a new issue if the problem persists with the latest release.

Issue #2722, opened 2020-05-30
[FEATURE] `npm install` script should be renamed


XhmikosR:
`install` is a built-in npm script. If this is meant to run as `postinstall` it should be changed to that.

REJack:
I agree with that, idk why I used `install` instead of `postinstall`.

PR #2727, opened 2020-05-30
Add stylelint support

XhmikosR:
@REJack this should supersede the https://github.com/ColorlibHQ/AdminLTE/tree/feat/stylelint branch.

I didn't add `fusv` because you first need to address #2725.

Waiting for #2721 and then I'll add `npm test` on CI too.

PS. better not squash the patches.
PPS. These changes will probably slightly improve gzip compression, at least for the non-minified CSS files. We use bundlewatch to monitor the sizes, feel free to add it or let me know and I could take a stab at it after there is a working CI.

Non-whitespace diff: https://github.com/ColorlibHQ/AdminLTE/pull/2727/files?w=1


.stylelintignore
```diff
@@ -0,0 +1,4 @@
+**/*.min.css
```
REJack, line 1: Why we ignore not any css files? I would only lint the scss files. 🤣

XhmikosR: Well, I leave the filetype unspecified. It will lint whatever files are supported. We do the same upstream too.

XhmikosR:
@REJack how do you want the CI to be? Running `npm test` on each matrix seems redundant so I was thinking to only run it on Ubuntu and Node.js 12.x.

REJack:
Yup `npm test` should only executed one time, not like the Node.js Install test 😄 ubuntu-latest with Node.js 12 sounds good for me.

.stylelintignore
REJack: Ok :D it was only a question :P

XhmikosR:
Actually, now that I think about it I could move lint to a separate workflow. I think it makes more sense. And keep ci.yml with deps + building.

REJack:
Yeah I would create seperate workflow for linting too.

XhmikosR:
@REJack WDYT?

XhmikosR:
BTW you really need to remove `install` because it runs on every `npm i`. See #2722.

.stylelintignore
XhmikosR: No worries :)

.stylelintrc
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": [
+    "stylelint-config-twbs-bootstrap/scss"
+  ],
+  "rules": {
+    "declaration-no-important": null,
+    "order/properties-order": null,
```
XhmikosR, line 7: I set all these to `null` but should be revisited later, which should help with #2742

XhmikosR:
Anyway feel free to merge @REJack if it looks good to you. But better not squash the patches :)

REJack:
I look later over this and the other PRs & Issues, I'm not at my Notebook right now 😄

XhmikosR:
I might need to adapt this after #2716 is merged, so I'll mark it as draft until that's merged.

.stylelintrc
REJack: > I set all these to null but should be revisited later, which should help with #2742.

I've added a note to the v3.1 Project to keep an eye on it 😄

XhmikosR:
@REJack please don't squash the patches when I say so :/ It's better for history to keep some patches separate.

This applies to this PR too.

REJack:
Should I merge my branch before merging this PR or it is not needed?

XhmikosR:
No, I can always fix merge conflicts and rebase myself. :)

PR #2728, opened 2020-05-30
Add ESLint support

XhmikosR:
Requires #2727.

Notes:

1. I've disabled a couple of rules because I don't know the codebase 100%, but should be enabled/revisited later:

* `eqeqeq`
* `no-eq-null` this you might want to skip if you want to loosely check for this, although better be explicit or just use `!foo`

2. I also made sure the docs JS are in ES5
3. I removed `babel-eslint` since it's not needed with the features you are using AFAICT
4. ~~Still pending to add `npm test` on CI due to #2721~~ Added
5 There are 2 failures that need to be fixed; one because `Number.isFinite` is used which isn't supported on IE 10 and one because in docs you are using `a.href` (https://eslint.org/docs/rules/no-script-url)

PS. These changes will probably slightly improve gzip compression, at least for the non-minified JS files.

Non-whitespace diff: https://github.com/ColorlibHQ/AdminLTE/pull/2728/files?w=1


build/js/TodoList.js
```diff
@@ -49,29 +48,29 @@ const TodoList = (($) => {
...
-      this._config.onUnCheck.call(item);
+    unCheck(item) {
+      this._config.onUnCheck.call(item)
     }
 
     // Private
 
     _init() {
-      var that = this
+      const that = this
```
XhmikosR, line 71: Need to double check if this is needed anymore :)

dist/js/demo.js
```diff
@@ -29,196 +32,196 @@
...
   })
-  var navbar_all_colors       = navbar_dark_skins.concat(navbar_light_skins)
-  var $navbar_variants_colors = createSkinBlock(navbar_all_colors, function (e) {
+  var navbar_all_colors = navbar_dark_skins.concat(navbar_light_skins)
+  var $navbar_variants_colors = createSkinBlock(navbar_all_colors, function () {
     var color = $(this).data('color')
     var $main_header = $('.main-header')
     $main_header.removeClass('navbar-dark').removeClass('navbar-light')
-    navbar_all_colors.map(function (color) {
+    navbar_all_colors.forEach(function (color) {
```
XhmikosR, line 224: `map` was wrong AFAICT since no return value was specified.

.stylelintignore
```diff
@@ -0,0 +1,4 @@
+**/*.min.css
```
REJack, line 1: Why we ignore not any css files? I would only lint the scss files.

REJack: Oh wrong PR 🤣

XhmikosR: This is unrelated to this patch though. See #2727. That is the branch for stylelint.

XhmikosR:
After we are done with the open PRs, I have a few JS tweaks to make.

XhmikosR:
This PR apart from requiring #2727, needs the JS lint issues tackled. I could use some help with this.

XhmikosR:
@REJack we need to get those 2 issues fixed. Feel free to fix them after the merge :)

Also, the code that skips dotfiles, will skip directories also that start with a dot. Feel free to improve it later.

REJack:
I will fix the 2 issues later this evening when I arrived at home 😄 also I will impove the filter to ignore html & typescript files.

XhmikosR:
Ping me when you do so in #2751 :)

XhmikosR:
@REJack forgot to mention, always be careful with what `unicorn` rules you enable because many times they target super modern browsers, which might not be what one wants. Hence why I have disabled many of them in this PR. Although eslint-plugin-compat should error, but better know about this :)

PR #2733, opened 2020-05-31
Use https

XhmikosR:
This will conflict with #2731, so I'm marking it as draft for now

### DIFF
--- a/pages/UI/modals.html
+++ b/pages/UI/modals.html
@@ -777,7 +777,7 @@
                 </button>
                 <div class="text-muted mt-3">
                   Instructions for how to use modals are available on the
-                  <a href="http://getbootstrap.com/javascript/#modals">Bootstrap documentation</a>
+                  <a href="https://getbootstrap.com/docs/4.5/components/modal/">Bootstrap documentation</a>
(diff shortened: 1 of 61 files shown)


pages/UI/modals.html
```diff
@@ -777,7 +777,7 @@ <h3 class="card-title">
                 </button>
                 <div class="text-muted mt-3">
                   Instructions for how to use modals are available on the
-                  <a href="http://getbootstrap.com/javascript/#modals">Bootstrap documentation</a>
+                  <a href="https://getbootstrap.com/docs/4.5/components/modal/">Bootstrap documentation</a>
```
XhmikosR, line 780: Unfortunately, we don't have a `latest` link upstream so I went with the latest version. Alternatively we could use `https://getbootstrap.com/components/` but that will redirect to `https://getbootstrap.com/docs/4.5/components/alert`

REJack: At the moment we doesn't use Bootstrap 4.5, we are on 4.4 😄 But I guess Bootstrap 4.5 works without any problems here.

PR #2740, opened 2020-05-31
Fix prefix script to run for all CSS files

XhmikosR:
I have to say, that was a nice bug 😁

So, basically, the split dist CSS files were not using proper CSS prefixes all this time.


XhmikosR:
I have to say, that was a nice bug 😁

So, basically, the split dist CSS files were not using proper CSS prefixes all this time.

REJack:
> So, basically, the split dist CSS files were not using proper CSS prefixes all this time.

🤦 That was my fault 😄

PR #2745, opened 2020-05-31
A few HTML validation fixes

XhmikosR:
The rest after I set up the validator and we are done with the open PRs :)

### DIFF
--- a/pages/examples/e-commerce.html
+++ b/pages/examples/e-commerce.html
@@ -748,7 +748,7 @@
               <h4>Available Colors</h4>
               <div class="btn-group btn-group-toggle" data-toggle="buttons">
                 <label class="btn btn-default text-center active">
-                  <input type="radio" name="color_option" id="color_option1" autocomplete="off" checked="">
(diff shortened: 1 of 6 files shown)


pages/examples/e-commerce.html
```diff
@@ -748,7 +748,7 @@ <h3 class="my-3">LOWA Men’s Renegade GTX Mid Hiking Boots Review</h3>
               <h4>Available Colors</h4>
               <div class="btn-group btn-group-toggle" data-toggle="buttons">
                 <label class="btn btn-default text-center active">
-                  <input type="radio" name="color_option" id="color_option1" autocomplete="off" checked="">
+                  <input type="radio" name="color_option" id="color_option1" autocomplete="off" checked>
```
XhmikosR, line 751: BTW this doesn't seem to have any visual indication that it's checked/active even though the parent label has the active class.

REJack: Right it has no visual indication its ok for the demo, but some of the user don't know about default HTML code they would not get the correct on value and would open a issue for this 🤣 I would stay with this in the code.

PR #2751, opened 2020-06-01
JS tweaks

XhmikosR:
Requires #2728 and thus #2727 merged.

Also, needs thorough review because I've changed the logic in some places to improve performance. I kept the patches separately so that it's easier to review.

Added caching of some vars, although it shouldn't make a huge difference, it's considered a good practice if one is reusing the same jQuery object, because we avoid querying the DOM multiple times. The are probably more cases, but this should be a good first step :)

Non-whitespace comparison: https://github.com/ColorlibHQ/AdminLTE/pull/2751/files?w=1

TODO:

- [x] Merge the dependent PRs; the lint failures should be fixed in #2728
- [x] Test, test, test
- [x] Maybe try to squash a couple of similar patches before the PR is ready for review

### DIFF
--- a/build/js/Layout.js
+++ b/build/js/Layout.js
@@ -93,21 +93,23 @@ const Layout = ($ => {
         }
       }
 
-      if ($('body').hasClass(ClassName.LAYOUT_FIXED)) {
-        if (offset !== false) {
-          $(Selector.CONTENT).css('min-height', (max + offset) - heights.header - heights.footer)
-        }
+      if (!$('body').hasClass(ClassName.LAYOUT_FIXED)) {
+        return
+      }
 
-        if (typeof $.fn.overlayScrollbars !== 'undefined') {
-          $(Selector.SIDEBAR).overlayScrollbars({
-            className: this._config.scrollbarTheme,
-            sizeAutoCapable: true,
-            scrollbars: {
-              autoHide: this._config.scrollbarAutoHide,
-              clickScrolling: true
-            }
-          })
-        }
+      if (offset !== false) {
+        $(Selector.CONTENT).css('min-height', (max + offset) - heights.header - heights.footer)
+      }
+
+      if (typeof $.fn.overlayScrollbars !== 'undefined') {
+        $(Selector.SIDEBAR).overlayScrollbars({
+          className: this._config.scrollbarTheme,
+          sizeAutoCapable: true,
+          scrollbars: {
+            autoHide: this._config.scrollbarAutoHide,
+            clickScrolling: true
+          }
+        })
       }
     }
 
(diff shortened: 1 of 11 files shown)


XhmikosR:
This can be reviews @REJack. Unfortunately I cannot clean up the patches more because there are many conflicts.

I tested things as good as I could and everything seems to be fine. But please confirm the logic change patches, i.e. you might not like moving some stuff in `_init()`.

BTW these changes shouldn't be breaking AFAICT.

REJack:
I'll review this PR later, I've to load my car and then drive home 😄

XhmikosR:
I kept only the safest patches. I'll move the other patches to another PR later at some point.

XhmikosR:
@REJack please check each patch separately to make sure I didn't miss anything and better not squash these :)

build/js/Layout.js
```diff
@@ -116,7 +116,7 @@ const Layout = ($ => {
     fixLoginRegisterHeight() {
       if ($(Selector.LOGIN_BOX + ', ' + Selector.REGISTER_BOX).length === 0) {
         $('body, html').css('height', 'auto')
-      } else if ($(Selector.LOGIN_BOX + ', ' + Selector.REGISTER_BOX).length !== 0) {
+      } else {
```
XhmikosR, line 119: `length` can either be `0` (element not present) or not `0` so this looked redundant

build/js/Layout.js
```diff
@@ -60,7 +60,7 @@ const Layout = ($ => {
     fixLayoutHeight(extra = null) {
       let controlSidebar = 0
 
-      if ($('body').hasClass(ClassName.CONTROL_SIDEBAR_SLIDE_OPEN) || $('body').hasClass(ClassName.CONTROL_SIDEBAR_OPEN) || extra === 'controlSidebar') {
+      if ($('body').hasClass(ClassName.CONTROL_SIDEBAR_SLIDE_OPEN) || $('body').hasClass(ClassName.CONTROL_SIDEBAR_OPEN) || extra === 'control_sidebar') {
```
XhmikosR, line 63: This isn't documented BTW

REJack: In my opinion it's not need to be documented, its only can used by the internal logic and can't used by users.

XhmikosR: Well, it's public API though. That's why I reverted that.

REJack: That's true, but it can't be used directly at the moment. It can only used with the jQuery interface (or I can’t do it any other way), this will come with the jQuery dependency free version 😄

XhmikosR: That's sort of unrelated, it can be used directly through the API right now and it will still be public API unless you change the visibility of the `fixLayoutHeight` function. Nothing to do with jQuery being used or not :)

REJack: Ok ^^ I have no idea how to document this 🤣 that's one of the things I hate.

PR #2752, opened 2020-06-01
Plugins

XhmikosR:


pages/forms/advanced.html
```diff
@@ -1372,7 +1372,7 @@ <h3 class="card-title">Bootstrap Switch</h3>
 <script src="../../plugins/bootstrap4-duallistbox/jquery.bootstrap-duallistbox.min.js"></script>
 <!-- InputMask -->
 <script src="../../plugins/moment/moment.min.js"></script>
-<script src="../../plugins/inputmask/min/jquery.inputmask.bundle.min.js"></script>
```
XhmikosR, line 1375: So all this time v4.0.9 was being used wrongfully :)

REJack: Oh dang, I know the bundle file includes already moment.js 🤦

XhmikosR: It doesn't seem to be a bundle file anymore?

XhmikosR: Might be worth pruning the plugins folder from time to time, but it needs manual review because of .npmignore and that flot-old folder.

REJack: > It doesn't seem to be a bundle file anymore?

It looks like the author removed the bundle file with v5.

> Might be worth pruning the plugins folder from time to time, but it needs manual review because of .npmignore and that flot-old folder.

That's a good Idea, at the latest with v4 I'll sort out a lot of plugins to remove jQuery as dependency.

PR #2754, opened 2020-06-01
Add bundlewatch

XhmikosR:
IMO it's useful, but you might not want to do it :)

That being said, the limits I chose are mostly strict. You can tweak them as needed. Also, this checks with gzip compression, but you can tweak it as needed. https://bundlewatch.io/#/reference/configuration

TODO:

- [x] Add `BUNDLEWATCH_GITHUB_TOKEN` in the repository secrets https://service.bundlewatch.io/setup-github

Note that this won't run from forks PRs due to the secrets not being available. But the npm script will run which should mark the build as failed if the limits are surpassed.

### DIFF
--- a/.bundlewatch.config.json
+++ b/.bundlewatch.config.json
@@ -1,0 +1,65 @@
+{
+  "files": [
+    {
+      "path": "./dist/css/adminlte.css",
+      "maxSize": "80 kB"
+    },
+    {
+      "path": "./dist/css/adminlte.min.css",
+      "maxSize": "72.5 kB"
+    },
+    {
+      "path": "./dist/css/alt/adminlte.components.css",
+      "maxSize": "14 kB"
+    },
+    {
+      "path": "./dist/css/alt/adminlte.components.min.css",
+      "maxSize": "13 kB"
+    },
+    {
+      "path": "./dist/css/alt/adminlte.core.css",
+      "maxSize": "46 kB"
+    },
+    {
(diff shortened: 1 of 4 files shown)


REJack:
Why we add this not in the lint workflow?

I've added the token in secrets 😄

XhmikosR:
> Why we add this not in the lint workflow?

Because we need to do a build in order to get the latest results :)

REJack:
Ok 😄

XhmikosR:
@REJack Are you sure you added the app and the secret to the repo properly? I don't see it running on master yet.

XhmikosR:
Ah, nvm actually it throws an error.

XhmikosR:
@REJack better remove the app from the repo, add it again, and add the secret again. I see this error currently:

```
 [ERROR] GitHubService HTTP_404 :: Not Found
```

REJack:
I can't remove it, I can only generate a new secret. There is not a integration like LGTM.com has.

XhmikosR:
Didn't you authorize the app? If not that's the reason for the failure and possible for LGTM too.

REJack:
Oh may it stuck on the group authorize request.
LGTM is integrated, this was authorized already on the group otherwise I could not integrate it.
lgtm integration

REJack:
The error is gone in my last PR 😄

XhmikosR:
Yeah, someone needs to authorize bundlewatch for the repo, re-add the secret. Also, ideally remove LGTM and re-add it.

PR #2763, opened 2020-06-01
Fix ESLint errors

REJack:


build/js/Layout.js
```diff
@@ -131,7 +131,7 @@ const Layout = ($ => {
 
       if (this._config.loginRegisterAutoHeight === true) {
         this.fixLoginRegisterHeight()
-      } else if (Number.isInteger(this._config.loginRegisterAutoHeight)) {
+      } else if (this._config.loginRegisterAutoHeight === parseInt(this._config.loginRegisterAutoHeight))) {
```
XhmikosR, line 134: Note that this isn't 100% the same https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#Polyfill

REJack: Do you think we should use the polyfill instead?

dist/js/demo.js
```diff
@@ -366,8 +366,9 @@
   })
   $container.append($logo_variants)
   var $clear_btn = $('<a />', {
```
XhmikosR, line 368: Could probably make this a proper `button` :)

REJack: I considered to remove it instead of fixing it 🤣 idk why it's only on logo variants.

build/js/Layout.js
XhmikosR: Your call, if you think this can fail with NaN. But putting the polyfill needs some thought, like having a utils file and so on.

XhmikosR: So, maybe just fix the new lint error (always specify the radix) and see how it goes. It's not 100% the same, though.

REJack: > Your call, if you think this can fail with NaN.

What you mean? 😪 I'm a bit tired 😄

REJack: I added the radix (10), the lint error is gone 😄 I'll test it tomorrow with any other change.

XhmikosR: https://stackoverflow.com/questions/14636536/how-to-check-if-a-variable-is-an-integer-in-javascript

But maybe using `parseInt` is OK in your case here.

REJack: That's the same stackoverflow that I've found 🤣 I guess it works for this case.

PR #2767, opened 2020-06-02
Clean up more plugins' files.

XhmikosR:


.gitignore
```diff
@@ -1,37 +1,59 @@
...
-// Plugins
-/plugins/*/*.md
-/plugins/*/*.html
-/plugins/*/package.json
-
-
-// Docs
+# Plugins
+/plugins/**/*.md
+!/plugins/**/LICENSE.md
```
XhmikosR, line 24: Those could be simplified with regex, but I don't have the time to look into it more.

Also, ideally the files shouldn't be copied in the first place.

PR #2772, opened 2020-06-02
Lint docs

XhmikosR:
Notes:

1. I didn't add the new docs-lint to `npm test` but later I will
2. I need some feeback how to proceed with the non-obvious errors like those `aria` ones


docs/faq.md
```diff
@@ -3,13 +3,14 @@ layout: page
...
-AdminLTE is an HTML template that can be used for any purpose. However, it is not made to be easily installed on Wordpress. It will require some effort and enough knowledge of the Wordpress script to do so.
+#### Can AdminLTE be used with WordPress?
+
+AdminLTE is an HTML template that can be used for any purpose. However, it is not made to be easily installed on WordPress. It will require some effort and enough knowledge of the WordPress script to do so.
 
 #### Is there an integration guide for PHP frameworks such as Yii or Symfony?
+
 Short answer, no. However, there are forks and tutorials around the web that provide info on how to integrate with many different frameworks. There are even versions of AdminLTE that are integrated with jQuery ajax, AngularJS and/or MVC5 ASP .NET.
 
-<!---
```
XhmikosR, line 12: I can remove this whole paragraph instead.

REJack: You mean the paragraph below? 😄

XhmikosR: Yeah

REJack: Yeah this can be removed :P it was only a old paragraph from the v2 docs 🤣

XhmikosR: Actually it's not a bad idea to keep it just mentioning GH releases.

XhmikosR:
There are 2 warnings left which might make sense to fix later. I didn't turn `Werror` on for now.

PR #2776, opened 2020-06-02
fix nav-link fadeIn

REJack:


XhmikosR:
@REJack it seems this introduced a regression: the `sidebar-is-opening` is not removed after the sidebar is visible.

XhmikosR:
Wrong, PR. Must have been introduced in #2775

PR #2788, opened 2020-06-03
Remove ionic icons from pages that don't use them

XhmikosR:


REJack:
You can remove it from index2.html too 😄 only index.html & index3.html use the IonIcons.

XhmikosR:
index2.html doesn't have the CSS?

REJack:
Oh you are right 🤦‍♂️ I thought it has the CSS too.

PR #2789, opened 2020-06-03
Use an object when setting multiple CSS properties

XhmikosR:


XhmikosR:
BTW I don't have any benchmark, the alternative would be to chain the calls.

REJack:
I've benchmarked it with [measurethat.net](https://www.measurethat.net/Benchmarks/Show/8459/0/jquery---css-multiple-vs-css-single-stack-vs-css-single) (Object vs Chained vs Single Rows). It looks like chained is faster then the other two.

XhmikosR:
Yeah, that's why I have suggested that most CSS is moved out of JS.

Either way, this patch should be the same performance-wise with what happens on master. Also, this is consistent with the rest of the codebase.

XhmikosR:
BTW I do plan to cache the selectors later, still splitting the patches

XhmikosR:
@REJack so can we move this so that I move to the rest of the perf JS improvement tweaks?

REJack:
> @REJack so can we move this 

Like merge it?

XhmikosR:
Yeah, typo, I meant move with this. :)

REJack:
Sure, I will prep the version after merging it to v3.1.0-pre 😄

PR #2790, opened 2020-06-04
Cache selectors when possible.

XhmikosR:
@REJack please test this carefully, in case I overdid it. I tested this, but the more the better.


REJack:
I'll test this later 😄 first I set on all v3.1.0 issues the `state:pending` tag 🤣 then I will eat something.

PR #2791, opened 2020-06-04
Refactor SCSS main files.

### LINKED ISSUE #2734: [FEATURE] run node-sass less times
It's marginally faster, see https://github.com/twbs/bootstrap/blob/master/package.json#L24

The only issue is the different output dir you have, but I guess you could run node-sass twice instead of 6 times :)

XhmikosR:
Fixes #2734

Notes:

- [x] If you have any docs that you mention how to extend the files, you should adapt the content
- [x] This should probably mentioned in the release notes in details; the file case has also changed


REJack:
> If you have any docs that you mention how to extend the files, you should adapt the content

We have no informations about extending the SCSS files, Idk how to write this 😄 

> This should probably mentioned in the release notes in details; the file case has also changed

I've added a mention in the release draft :D

PR #2794, opened 2020-06-04
Layout.js: minor simplification

XhmikosR:
I think that was the original intention, but please confirm :)

### DIFF
--- a/build/js/Layout.js
+++ b/build/js/Layout.js
@@ -75,15 +75,12 @@ class Layout {
     }
 
     const max = this._max(heights)
-    let offset = this._config.panelAutoHeight
-
-    if (offset === true) {
-      offset = 0
-    }
+    const { panelAutoHeight } = this._config
+    const offset = panelAutoHeight === true ? 0 : offset
 
     const $contentSelector = $(Selector.CONTENT)
 
-    if (offset !== false) {
+    if (panelAutoHeight !== false) {
       if (max === heights.controlSidebar) {
         $contentSelector.css('min-height', (max + offset))
       } else if (max === heights.window) {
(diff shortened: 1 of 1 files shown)


XhmikosR:
BTW there's still a flashing in the sidebar items on page load on master.

build/js/Layout.js
```diff
@@ -75,15 +75,12 @@ class Layout {
     }
 
     const max = this._max(heights)
-    let offset = this._config.panelAutoHeight
-
-    if (offset === true) {
-      offset = 0
-    }
+    const { panelAutoHeight } = this._config
+    const offset = panelAutoHeight === true ? 0 : offset
```
REJack, line 79: This should look like this
```js
const offset = panelAutoHeight === true ? 0 : panelAutoHeight
```

Otherwise it triggers only `false` if there is int set instead of a boolean 😄

XhmikosR: Err typo, good catch!

XhmikosR: Actually, it's not a typo. I tried to separate the offset variable from the `panelAutoHeight` one. I still use the `offset` variable below.

XhmikosR: NVM, I need more coffee, please check again.

XhmikosR: OK, third time's a charm. My initial patch was right. In this patch I split the variables. So, `offset` holds the value of `offset` and `panelAutoHeight` is a Boolean. If `panelAutoHeight` can get a Boolean, this should be cleared up more.

REJack: The problem is that it will not change the offset with your inital patch, it results everytime to 0 if it's set to a positive number or negative number, false & true works like it should.

XhmikosR: I don't see how it works, when the strict comparison was already being used. If you want to accept Boolean values we should cast the variable to Boolean.

XhmikosR: OK, I see now. `panelAutoHeight` accepts both a Boolean and a Number. This is why we have basic options type checking upstream. I'll see if this patch can be adapted otherwise I'll drop it.

PR #2796, opened 2020-06-05
Update devDependencies and gems

XhmikosR:
1. when you bump the version don't forget package-lock.json.
2. it might make more sense to bump the version before a release instead of randomly :)
3. might be worth looking into adding dependabot and configuring it to run weekly or monthly for the docs gems and the npm deps instead of doing this manually. It's not a perfect solution but should automate big part of the process


REJack:
> 2. it might make more sense to bump the version before a release instead of randomly :)

I only bump the version at the begin of a new version as `-pre` to show the user that they use a pre-version and not a full release. I only bumped again since I decided to release v3.1.0 as next version 😄 

> 3. might be worth looking into adding dependabot and configuring it to run weekly or monthly for the docs gems and the npm deps instead of doing this manually. It's not a perfect solution but should automate big part of the process

How I can setup the dependabot for this?

XhmikosR:
Dependabot is an app owned by GitHub. You can add it to the repo, but configure it properly so that it doesn't check too frequently. Maybe weekly or monthly. And I'm not sure if you should allow it to check for dependencies since the git files need to be updated too.

PR #2797, opened 2020-06-05
Fix docs

### LINKED ISSUE #2792: [BUG] Fix docs baseurl
I could try having a stab at it later, what's the final URL you will use? You might be missing `baseurl: /docs/3.0` in _config.yml

XhmikosR:
Sorry for not splitting these, but it's hard. Just check the non-whitespace diff which is cleaner: https://github.com/ColorlibHQ/AdminLTE/pull/2797/files?w=1

Fixes #2792

EDIT: this might break local browsing of the generated docs pages. I'll verify that later, but it might be better dropping docs_html from the npm/zip packages.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -2,11 +2,9 @@ title: AdminLTE v3 Documentation
 version: v3.1.0-pre
 description: AdminLTE v3 Documentation
 
-markdown: kramdown
-kramdown:
-  auto_ids: false
+baseurl: /docs/3.1
+url: https://adminlte.io
 
-highlighter: rouge
 plugins:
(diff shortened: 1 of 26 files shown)


docs/_config.yml
```diff
@@ -2,11 +2,9 @@ title: AdminLTE v3 Documentation
 version: v3.1.0-pre
 description: AdminLTE v3 Documentation
 
-markdown: kramdown
```
XhmikosR, line 5: These are the defaults for Jekyll 4.x, hence why I removed them.

XhmikosR:
I went ahead and removed `docs_html` from the npm package.

PR #2799, opened 2020-06-05
HTML Pages tag cleanup

XhmikosR:


XhmikosR:
BTW personally I prefer lowercase `doctype` because this is HTML not XHTML, and offers slightly better compression.

PR #2818, opened 2020-06-06
vnu: enable `Werror`

XhmikosR:
Refs https://github.com/ColorlibHQ/AdminLTE/pull/2528#issuecomment-639997251

Feel free to push to this branch to fix the table warnings. The rest I guess we could ignore them, although they are still valid (you probably don't want to use a `section` there). The problem is we can't ignore only a specific line, we need to disable the error/warning globally.

So, ideally, all warnings should be fixed so that we don't have have to ignore any extra stuff. :)

### DIFF
--- a/build/js/ExpandableTable.js
+++ b/build/js/ExpandableTable.js
@@ -48,7 +48,7 @@ class ExpandableTable {
   init() {
     $(Selector.HEADER).each((_, $header) => {
       // Next Child to the header will have the same column span as header
-      $($header).next().children().first().attr('colSpan', $($header).children().length)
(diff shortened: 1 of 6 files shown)


XhmikosR:
@REJack I had a closer look. It seems the docs warnings were valid, I also found a broken link and fixed all that.

About the ExpandableTable `colspan`: I see the JS plugin adds them but I'm not sure if it should. Because this way we promote proper markup plus it's one less thing to do in JS.

That being said, I suppose you checked if the whole Expandable tables feature was possible with the upstream collapse plugin?

PR #2823, opened 2020-06-08
Fix accordion example

XhmikosR:
* make collapsible heading behave like a block
* remove the non-existent `panel-collapse` class
* replace the `in` class with `show`


REJack:
The row is not correct :)
preview

XhmikosR:
Well, fee to fix it :)

XhmikosR:
NVM, it was easier than I though. BTW this is an example that the accessibility upstream suggestions are not followed ;)

REJack:
> BTW this is an example that the accessibility upstream suggestions are not followed ;)

I would say we have a lot of these examples 🤣

PR #2825, opened 2020-06-08
rework ExpandableTable

REJack:


XhmikosR:
@REJack please check #2818 and my last comment there.

REJack:
> About the ExpandableTable colspan: I see the JS plugin adds them but I'm not sure if it should. Because this way we promote proper markup plus it's one less thing to do in JS. EDIT: also the page won't jump if colspan is set via HTML. I suggest that you require proper colspan via HTML.

I removed the colspan setup via JS and placed it directly in the HTML.

> That being said, I suppose you checked if the whole Expandable tables feature was possible with the upstream collapse plugin?

It would work but it doesn't slide down/up the content, it only hides/shows the content.

XhmikosR:
Wrong PR, but I get the gist :P

PR #2834, opened 2020-06-13
Be explicit with the background color property

### LINKED ISSUE #2726: [BUG] Revisit using `background`
It seems `background` is used in many places where only `background-color` is meant to be changed. It's better to be explicit.

The same might happen for other shorthands, I didn't check everything.

XhmikosR:
Notes:

1. I didn't change the scss/plugins files because some plugins do use the shorthand. **It should be tackled too, though, by using `background-image` or `background-color` where appropriate.**

@REJack feel free to push to this branch to fix the plugin instances :)

Fixes #2726

### DIFF
--- a/build/scss/_alerts.scss
+++ b/build/scss/_alerts.scss
@@ -26,7 +26,7 @@
 @each $color, $value in $theme-colors {
   .alert-#{$color} {
     color: color-yiq($value);
-    background: $value;
+    background-color: $value;
     border-color: darken($value, 5%);
   }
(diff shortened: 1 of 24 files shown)


XhmikosR:
@REJack you missed my comment about the plugins, or will you handle that later? Do note that you can push to a PR's branch.

REJack:
I've added a note on the v3.1 board for it and keep an eye on it 😄 we can change it later.

PR #2835, opened 2020-06-13
Use template literals

XhmikosR:
Draft because the other PRs need to be sorted.

### DIFF
--- a/build/js/CardWidget.js
+++ b/build/js/CardWidget.js
@@ -67,7 +67,7 @@ class CardWidget {
         this._parent.addClass(CLASS_NAME_COLLAPSED).removeClass(CLASS_NAME_COLLAPSING)
       })
 
-    this._parent.find('> ' + SELECTOR_CARD_HEADER + ' ' + this._settings.collapseTrigger + ' .' + this._settings.collapseIcon)
+    this._parent.find(`> ${SELECTOR_CARD_HEADER} ${this._settings.collapseTrigger} .${this._settings.collapseIcon}`)
       .addClass(this._settings.expandIcon)
       .removeClass(this._settings.collapseIcon)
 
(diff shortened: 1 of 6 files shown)


build/js/CardWidget.js
```diff
@@ -125,11 +125,10 @@ class CardWidget {
   }
 
   minimize() {
-    this._parent.find(this._settings.maximizeTrigger + ' .' + this._settings.minimizeIcon)
+    this._parent.find(`${this._settings.maximizeTrigger} .${this._settings.minimizeIcon}`)
       .addClass(this._settings.maximizeIcon)
       .removeClass(this._settings.minimizeIcon)
-    this._parent.css('cssText', 'height:' + this._parent[0].style.height + ' !important;' +
-      'width:' + this._parent[0].style.width + ' !important; transition: all .15s;'
+    this._parent.css('cssText', `height: ${this._parent[0].style.height} !important; width: ${this._parent[0].style.width} !important; transition: all .15s;`
```
XhmikosR, line 131: BTW this seems wrong and should be broken already

REJack: What seems wrong, the entire cssText row or what? 😄

XhmikosR: That's not how you set CSS with jQuery... See #2838

REJack: That’s the only way to add !important with jQuery's css(), you can take a look [here](https://makitweb.com/how-to-add-important-to-css-property-with-jquery/#csstext). Otherwise we need to use attr() for this 😄.

XhmikosR: I already split the patch so that we don't pollute unrelated PRs. We can discuss and tackle this there after the other PRs are done.

PR #2837, opened 2020-06-13
ControlSidebar.js: minor simplification

XhmikosR:
It's not a lot more readable, but at least the variables are set once. This could be simplified later further.


REJack:
I've added conflicts to your PR 🤣

PR #2838, opened 2020-06-13
CardWidget: fix `minimize()` CSS

XhmikosR:
Draft because it depends on the other PRs

### DIFF
--- a/build/js/CardWidget.js
+++ b/build/js/CardWidget.js
@@ -17,38 +17,32 @@ const DATA_KEY = 'lte.cardwidget'
 const EVENT_KEY = `.${DATA_KEY}`
 const JQUERY_NO_CONFLICT = $.fn[NAME]
 
-const Event = {
-  EXPANDED: `expanded${EVENT_KEY}`,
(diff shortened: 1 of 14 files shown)


build/js/CardWidget.js
```diff
@@ -128,8 +128,11 @@ class CardWidget {
     this._parent.find(`${this._settings.maximizeTrigger} .${this._settings.minimizeIcon}`)
       .addClass(this._settings.maximizeIcon)
       .removeClass(this._settings.minimizeIcon)
-    this._parent.css('cssText', `height: ${this._parent[0].style.height} !important; width: ${this._parent[0].style.width} !important; transition: all .15s;`
-    ).delay(10).queue(function () {
+    this._parent.css({
```
XhmikosR, line 131: I'm not sure if this is even used, it's just that before it didn't do anything

XhmikosR: It seems I was wrong and `cssText` is needed for `!important`.

I will investigate this further.

PR #2841, opened 2020-06-13
Sidebar Search

REJack:
## Sidebar Search

This PR brings a complete new JS Plugin called Sidebar Search,
with this Plugin you can search through your sidebar items and find the correct ones, including path to the items itself.

grafik grafik grafik

#### ToDo
- [x] Search Highlighting
- [x] Docs Section

@GerryGe Feel free to test it 😄 

Closes #2472

PR #2844, opened 2020-06-16
AdminLte Template 51 changed files

Biggies178:


REJack:
Copy of #2841 without changes

PR #2986, opened 2020-08-28
Fixed misaligned grids of main content

qudusG:
The four grids (2 grids per row) in the starter page were not properly aligned. When a new item is added to a grid in a row and not added to the 2nd grid in the same row, the row becomes misaligned. In short, the row becomes misaligned if the contents in one gird of the row is not the same size as the other grid of the same row. To fix this, I have assigned each grid it's own column dimension (col-lg-6) and eliminated col-md-6s which group two vertical grids horizontally from different rows. Grids are now correctly aligned.


REJack:
You used a old starter.html version, please use the lastest version from master otherwise I can't merge these changes.

PR #3159, opened 2020-11-26
sidebar upgradeed

danny007in:
Removed Unnecessary Code
And
Sidebar improved (Can work smoothly without overlayScrollbars plugin)


danny007in:
merge this PR, otherwise
Because if layout-fixed, without overlayScrollbars  plugin, sidebar will not scroll.
I have fixed in this PR

PR #3183, opened 2020-12-01
Post css  devdep update

danny007in:
@XhmikosR In package.json also script command


package.json
```diff
@@ -132,7 +132,7 @@
     "@babel/core": "^7.12.9",
     "@babel/preset-env": "^7.12.7",
     "@rollup/plugin-babel": "^5.2.1",
-    "autoprefixer": "^9.8.6",
+    "autoprefixer": "^10.0.2",
```
XhmikosR, line 135: There's a newer version of autoprefixer at the time of writing.

package.json
```diff
@@ -14,15 +14,15 @@
     "css-compile-splits": "npm run css-compile-bash -- build/scss/parts -o dist/css/alt/",
     "css-prefix": "postcss --config build/config/postcss.config.js --replace \"dist/css/*.css\" \"!dist/css/*.min.css\"",
     "css-prefix-splits": "postcss --config build/config/postcss.config.js --replace \"dist/css/alt/*.css\" \"!dist/css/alt/*.min.css\"",
-    "css-minify-bash": "cleancss --level 1 --format breakWith=lf --source-map --source-map-inline-sources --output ",
+    "css-minify-bash": "cleancss -O1 --format breakWith=lf --source-map --source-map-inline-sources --output ",
```
XhmikosR, line 17: This change and the `--rd` addition shouldn't be in this branch. They should be separate, clear patches.

XhmikosR (changes requested):

PR #3187, opened 2020-12-03
permit child indent in sidebar-collapse when focus

paul45:
permit child indent in sidebar-collapse when focus

### DIFF
--- a/dist/css/adminlte.css
+++ b/dist/css/adminlte.css
@@ -11461,7 +11461,7 @@ body.sidebar-collapse:not(.sidebar-mini-md):not(.sidebar-mini) .main-header::bef
   margin-left: -250px;
 }
 
-.sidebar-collapse .main-sidebar .nav-sidebar.nav-child-indent .nav-treeview {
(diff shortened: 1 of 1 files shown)


danny007in:
> 
> 
> permit child indent in sidebar-collapse when focus

Edit scss files not in css.
First check readme, before contribute 😄

paul45:
ok sorry

I am not good with scss 

what du you think about the modification 

maybe someone could rewrite in scss

danny007in:
> 
> 
> ok sorry
> 
> I am not good with scss
> 
> what du you think about the modification
> 
> maybe someone could rewrite in scss

By adding this feature, user interface may not be good 🤔

REJack:
I see no difference with this change, may a Video or some Screenshots would help to know whats your result of the change.

Also @danny007in is right, you should never change css files here caused that they get rebuild from the scss files every release.

SCSS is not witchcraft, it's basicly CSS with some extra features like nesting, functions & variables.

PR #3284, opened 2021-01-05
Preloader Added

danny007in:


REJack:
works nice, I would only add a dark-mode override for it.

danny007in:
> 
> 
> works nice, I would only add a dark-mode override for it.

You mean, Ur adding or I needed to add

REJack:
I will add a small darkmode overwrite, but you should add a docs section for it.

index.html
```diff
@@ -29,6 +29,11 @@
 <body class="hold-transition sidebar-mini layout-fixed">
 <div class="wrapper">
 
+  <!-- Preloader -->
+  <div class="preloader">
+    <img src="dist/img/AdminLTELogo.png" alt="AdminLTELogo" height="60">
```
XhmikosR, line 34: Always add width too

REJack:
@danny007in Add a docs section for the preloader then I will merge it, so we can publish the v3.1.0.

danny007in:
Is dark mode preloader added or not

REJack:
No I have no write access to your PR, I will add it after merge.

danny007in:
u can try now, i have allow from earlier

REJack:
I've added the darkmode, now you can add the docs section 😄

danny007in:
ok i will check doc properly and update it

danny007in:
and u needed to remove unwanted construct in TODO at dropdown.js

REJack:
> and u needed to remove unwanted construct in TODO at dropdown.js

ignore it

buncis:
did it mean to fix this bug? https://github.com/ColorlibHQ/AdminLTE/discussions/3319

and also if that is true how about adding preloader to starter page too ?

PR #3308, opened 2021-01-13
Update CI config

XhmikosR:
* remove `CI: true` since it's set by the runner
* update to `actions/setup-node@v2`
* switch to `ruby/setup-ruby` Action
* specify `FORCE_COLOR: 2`
* lowercase `runner.os`
* update to Ruby 2.7
* update to Node.js 14.x
* fix npm cache mismatches due to wrong variable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
       - "**"
 
 env:
-  CI: true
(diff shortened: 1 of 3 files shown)


danny007in (approved):
i am not 100% sure its perfect, 
but seeing this code, i think there is no conflict 😄

PR #3309, opened 2021-01-13
Add CodeQL

XhmikosR:
I excluded the plugins folders because otherwise it takes too long to scan. That being said, I suggest that you scan everything in a temp branch (for example https://github.com/ColorlibHQ/AdminLTE/security/code-scanning?page=1&query=ref%3Arefs%2Fheads%2Fcodeql-test), report any issues upstream and/or consider using other plugins if there are replacements.

Obviously you should wait a bit for the upstream developers to have a look at these issues because some might be false positives :)

PS. I already contacted the OverlayScrollbars dev so that's one less project to contact

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,5 @@
+name: "CodeQL config"
+
+paths-ignore:
+  - docs/assets/plugins/
+  - plugins/
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,35 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches:
+      - master
+      - "!dependabot/**"
+  pull_request:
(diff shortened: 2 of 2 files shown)


REJack:
> report any issues upstream and/or consider using other plugins if there are replacements.

With v4.0.0 we will replace/remove many Plugins since I will drop the jQuery support so all jquery based plugins will removed if no replacement is found or it has an own dependency less version.

XhmikosR:
I still think you should report the issues upstream, because they might not be aware of any potential issues.

Also, do not forget that the current patch doesn't scan the plugins because it's a lot slower. So, you will need to make a temp branch like I did above to check the plugins.

danny007in (approved):

REJack:
@XhmikosR can I disable LGTM now with this PR merged?

XhmikosR:
They are not 100% the same, so better keep both IMHO.

REJack:
Alright 👌🏼

PR #3347, opened 2021-01-24
Update and rename feature-request-for-adminlte-v3-x.md 

danny007in:


REJack:
In the order that we start with AdminLTE v4, we don't add any new features to v3.1.0 so I changed it to v4.x.

PR #3361, opened 2021-01-26
implemented lockfile-lint devDep

danny007in:
followed by https://github.com/twbs/bootstrap/blob/6ffb0b48e455430f8a5359ed689ad64c1143fac2/package.json#L60


danny007in:
Plz review perfectly, because i just followed bs4

PR #3455, opened 2021-02-24
fix sidebar light text color

danny007in:


danny007in:
bundlewatch issue @REJack just bump it.

REJack:
increase the core.min.css size limit about 0.3kb

PR #3492, opened 2021-03-09
Enhance: dark mode

danny007in:
> Plz take deep review


REJack:
Overall the Colors look good but the Blue is for me too dark, may you can replace $blue-alt with `#3f6791`.
And the Button (Marked Green) is to light, may you can check this. 😄

Screenshot_2021-03-12 AdminLTE 3 Dashboard

PR #3511, opened 2021-03-15
update docs

VanhanenJLA:
"404 Couldn't find the requested release version 3.1."


danny007in:
better to make change
- create both 3.1.0-rc and 3.0.5

danny007in (changes requested):
divide into two group stable and unstable

danny007in (approved):

PR #3524, opened 2021-03-18
weekly to monthly schedule of dependabot

danny007in:
Because to release AdminLTE v3.1.0 fast

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,8 +10,7 @@ updates:
   - package-ecosystem: bundler
     directory: "/docs/"
     schedule:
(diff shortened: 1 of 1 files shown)


redstonedesigner:
I disagree with this change - it limits the frequency of updates to dependencies, which could include important feature / security updates.
Why is this change needed in order to release 3.1.0 faster?

danny007in:
> 
> 
> I disagree with this change - it limits the frequency of updates to dependencies, which could include important feature / security updates.
> Why is this change needed in order to release 3.1.0 faster?

Some how ur right

PR #3540, opened 2021-03-26
Support SVG Font Awesome icon in main sidebar

### LINKED ISSUE #3532: [BUG] i tag in sidebar menu changed to SVG by fontawesome meaning navigation item with sub item don't have the arrow rotating once opened.
**Describe the bug**
AdminLTE sidebar css has the following:
https://github.com/ColorlibHQ/AdminLTE/blob/v3/build/scss/_main-sidebar.scss#L109
with the purpose to rotate the arrow appearing on the right hand side of a menu item when there is sub navigations and the menu is open. 
For me this arrow never rotated. The reason is because Font awesome translate i tags to SVG automatically. Although you can desactivate this by setting `autoReplaceSvg: false` in font awesome configuration, people may not know about it and therefore not set this. Also I think adminLTE should support this, and the current css doesn't. This is possibly a feature request, but none the less would be nice and quick to add / fix.

**To Reproduce**
Unsure, this was after a fresh installation using yarn and a ruby on rails application.

**Expected behavior**
I would expect the arrow to rotate and pointing down (have the class `fa-angle-down` instead of `fa-angle-left`) when I click on the navigation item to display it's sub navigation.

**Screenshots**
Menu closed
Screenshot 2021-03-24 at 18 39 43
Menu opened
Screenshot 2021-03-24 at 18 39 54
i tag in the code
Screenshot 2021-03-24 at 18 51 06

translated automatically once rendered
Screenshot 2021-03-24 at 18 32 31

**Environment (please complete the following information):**
- AdminLTE Version: 3.1
- Operating System:  macOS Big Sur
- Browser (Version): Chrome Version 89.0.4389.82 (Official Build) (x86_64)

**Additional context**
I think we can either add the following to _main-sidebar.scss (have tried that locally by adding the following in a custom css file and that fixed the problem):

```css
.menu-open,
  .menu-is-opening {
    > .nav-link {
      svg.right {
        @include rotate(-90deg);
      }
    }
  }
```
Alternatively we could not restrict the `i` tag in the current css, although I am unsure if there will be any side effect from doing that.

I happy do raise a PR for it.

liljack:
This commits adds support for SVG Font Awesome icon within the main sidebar. Font Awesome translate i tags to svg tags, resulting in the angle icon not rotating in the sidebar.

Fixes #3532

### DIFF
--- a/build/scss/_main-sidebar.scss
+++ b/build/scss/_main-sidebar.scss
@@ -105,6 +105,7 @@
   .menu-open,
   .menu-is-opening {
     > .nav-link {
+      svg.right,
       i.right {
         @include rotate(-90deg);
       }


liljack:
The Lint check failing is unrelated to my change. Assuming this was failing before 🤷‍♂️

danny007in:
send to master branch

liljack:
Meh, Friday night, my brain is not all there, could have written:
```css
i.right, svg.right {
  @include rotate(-90deg);
}
```
instead of what I submitted here. Would you like me to change that or happy with the current ?

build/scss/_main-sidebar.scss
```diff
@@ -108,6 +108,9 @@
       i.right {
```
danny007in, line 108: add here 
like
```scss
svg.right,
i.right {
```

build/scss/_main-sidebar.scss
```diff
@@ -108,6 +108,9 @@
...
         @include rotate(-90deg);
       }
+      svg.right {
+        @include rotate(-90deg);
+      }
```
danny007in, line 113: remove here

danny007in (approved):

liljack:
Happy for me to merge ?

PR #3555, opened 2021-03-31
add errorhandling to cardrefresh plugin

equada:
Currently, if a server error occurs while loading contents in the card-refresh plugin, the overlay just stays and the spinner turns forever. With this change, the overlay will be hidden if an error occurs and there is an error handler function in the settings

### DIFF
--- a/build/js/CardRefresh.js
+++ b/build/js/CardRefresh.js
@@ -36,6 +36,7 @@ const Default = {
   loadOnInit: true,
   responseType: '',
   overlayTemplate: '<div class="overlay"><i class="fas fa-2x fa-sync-alt fa-spin"></i></div>',
+  errorTemplate: '<span class="text-danger"></span>',
   onLoadStart() {},
   onLoadDone(response) {
     return response
@@ -76,6 +77,9 @@ class CardRefresh {
       this._removeOverlay()
     }, this._settings.responseType !== '' && this._settings.responseType).fail((jqXHR, textStatus, errorThrown) => {
       this._removeOverlay()
+      let msg = $(this.__settings.errorTemplate).text(errorThrown)
+      this._parent.find(this._settings.content).empty().append(msg)
+      
       this._settings.onLoadFail.call($(this), jqXHR, textStatus, errorThrown)
     })
 
--- a/build/js/CardRefresh.js
+++ b/build/js/CardRefresh.js
@@ -34,6 +34,7 @@ const Default = {
   content: '.card-body',
   loadInContent: true,
   loadOnInit: true,
+  loadErrorTemplate: true,
   responseType: '',
   overlayTemplate: '<div class="overlay"><i class="fas fa-2x fa-sync-alt fa-spin"></i></div>',
   errorTemplate: '<span class="text-danger"></span>',
@@ -78,8 +79,11 @@ class CardRefresh {
     }, this._settings.responseType !== '' && this._settings.responseType)
     .fail((jqXHR, textStatus, errorThrown) => {
       this._removeOverlay()
-      const msg = $(this._settings.errorTemplate).text(errorThrown)
-      this._parent.find(this._settings.content).empty().append(msg)
+
+      if (this._settings.loadErrorTemplate) {
+        const msg = $(this._settings.errorTemplate).text(errorThrown)
+        this._parent.find(this._settings.content).empty().append(msg)
+      }
 
       this._settings.onLoadFail.call($(this), jqXHR, textStatus, errorThrown)
     })
--- a/build/js/CardRefresh.js
+++ b/build/js/CardRefresh.js
@@ -75,9 +75,10 @@ class CardRefresh {
 
       this._settings.onLoadDone.call($(this), response)
       this._removeOverlay()
-    }, this._settings.responseType !== '' && this._settings.responseType).fail((jqXHR, textStatus, errorThrown) => {
+    }, this._settings.responseType !== '' && this._settings.responseType)
+    .fail((jqXHR, textStatus, errorThrown) => {
       this._removeOverlay()
-      const msg = $(this.__settings.errorTemplate).text(errorThrown)
+      const msg = $(this._settings.errorTemplate).text(errorThrown)
       this._parent.find(this._settings.content).empty().append(msg)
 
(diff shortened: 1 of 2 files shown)


danny007in:
resolve js-lint 🕵🏼

equada:
@danny007in

The html:
```html
<div class="card">
     <div class="card-header">
       <h5 class="card-title">This is a title</h5>
       <button type="button" class="btn btn-primary btn-sm float-right" data-card-widget="card-refresh" data-source="/public/ajax/route"><i class="fas fa-redo"></i> refresh</button>
     </div>
     <div class="card-body">

     </div>
</div>
```
When opening the page, the card starts loading:
image

Then, if for example an error occurs on the server-side (the route `/public/ajax/route`), or basically if the jquery ajax `$.get` request fails in any way, the card is still loading, the overlay stays, eventhough the request is already finished (failed) in the background:

image

The user gets no feedback that something went wrong and the button is behind the overlay. (I know this could be fixed with css z-index but thats not the point).

My fix hides the overlay if something went wrong and gives the developer the opportunity to implement a proper error-handling, which is currently not possible at all

danny007in:
Along with this,
 add background gray to red,
Which represents - Something went wrong.

equada:
@danny007in I'm not quite sure what you mean by this. Sure, the user should get a feedback in some form like I mentioned, but if something went wrong,  the overlay is now hidden to reveal the button (so I cant color the overlay red)

Coloring any other component of the card red, like the card, card-header, card-body or button seems invasive. The Developer should not fear that admin-lte randomly colors something red or gives classes to elements.

My suggestion would be to clear the card-body and put the `errorThrown` inside it, like this:
```js
this._parent.find(this._settings.content).html('<span class="text-danger">'+errorThrown+'</span>')
```

image

Would that be fine?

build/js/CardRefresh.js
```diff
@@ -73,7 +75,13 @@ class CardRefresh {
 
       this._settings.onLoadDone.call($(this), response)
       this._removeOverlay()
-    }, this._settings.responseType !== '' && this._settings.responseType)
+    }, this._settings.responseType !== '' && this._settings.responseType).fail((jqXHR, textStatus, errorThrown) => {
+      this._removeOverlay()
+      const msg = $(this.__settings.errorTemplate).text(errorThrown)
+      this._parent.find(this._settings.content).empty().append(msg)
+
+      this._settings.onLoadFail.call($(this), jqXHR, textStatus, errorThrown)
```
danny007in, line 83: why `onLoadFail`.
`.fail` is enough

build/js/CardRefresh.js
```diff
@@ -73,7 +75,13 @@ class CardRefresh {
 
       this._settings.onLoadDone.call($(this), response)
       this._removeOverlay()
-    }, this._settings.responseType !== '' && this._settings.responseType)
+    }, this._settings.responseType !== '' && this._settings.responseType).fail((jqXHR, textStatus, errorThrown) => {
```
danny007in, line 78: `,fail` nextliine

build/js/CardRefresh.js
```diff
@@ -73,7 +75,13 @@ class CardRefresh {
...
+      this._removeOverlay()
+      const msg = $(this.__settings.errorTemplate).text(errorThrown)
+      this._parent.find(this._settings.content).empty().append(msg)
```
danny007in, line 81: try to make single line

danny007in (changes requested):

build/js/CardRefresh.js
danny007in: also add `.done`

build/js/CardRefresh.js
```diff
@@ -73,7 +75,13 @@ class CardRefresh {
 
       this._settings.onLoadDone.call($(this), response)
       this._removeOverlay()
-    }, this._settings.responseType !== '' && this._settings.responseType)
+    }, this._settings.responseType !== '' && this._settings.responseType).fail((jqXHR, textStatus, errorThrown) => {
+      this._removeOverlay()
+      const msg = $(this.__settings.errorTemplate).text(errorThrown)
```
danny007in, line 80: `__settings` to `_settings`

build/js/CardRefresh.js
equada: because the other event functions are called `onLoadStart` and `onLoadDone` so it seemed logical to me to call it `onLoadFail` ?

danny007in: remove on 0nLoadFail because by default its reponse is same.
but not as loadStart or loadDone.

REJack:
@danny007in is this ready to merge too?

build/js/CardRefresh.js
danny007in: @REJack does onLoadFail is required or not

build/js/CardRefresh.js
```diff
@@ -36,10 +36,12 @@ const Default = {
   loadOnInit: true,
   responseType: '',
   overlayTemplate: '<div class="overlay"><i class="fas fa-2x fa-sync-alt fa-spin"></i></div>',
+  errorTemplate: '<span class="text-danger"></span>',
```
danny007in, line 39: add
`showErrorTemplate: true`

danny007in, line 39: add docs to it

build/js/CardRefresh.js
```diff
@@ -74,6 +76,13 @@ class CardRefresh {
       this._settings.onLoadDone.call($(this), response)
       this._removeOverlay()
     }, this._settings.responseType !== '' && this._settings.responseType)
+    .fail((jqXHR, textStatus, errorThrown) => {
+      this._removeOverlay()
+      const msg = $(this._settings.errorTemplate).text(errorThrown)
```
danny007in, line 81: if statement `showErrorTemplate`

danny007in (changes requested):

build/js/CardRefresh.js
REJack: I would stay with it, then you can trigger alerts or something else with this.

danny007in (approved):

equada:
@danny007in Thank you for completing the requested changes!

Issue #3556, opened 2021-04-01
[BUG] URLs with characters '?' and '=' not encoded properly in sidebar search results.


zammitjohn:
**Describe the bug**
URLs with characters '?' and '=' not text encoded properly in sidebar search results. 
? -> %3F
= -> %3D
 
**To Reproduce**
1. Add element to sidebar element with 'href' containing a '?' or '='
2. Search for the element from the sidebar
3. Click on the result

**Environment (please complete the following information):**
- AdminLTE Version: v3.1.0
- Browser (Version): Chrome (Latest)

danny007in:
add url here ill check

danny007in:
Fixed in https://github.com/ColorlibHQ/AdminLTE/pull/3557

zammitjohn:
Thanks @danny007in!

muhanadali100:
not working with me and i use the last update

danny007in:
> 
> 
> not working with me and i use the last update

did u compiled

muhanadali100:
> 
> 
> > not working with me and i use the last update
> 
> did u compiled

yep i complied and not working

PR #3557, opened 2021-04-02
Fix: sidebar search url issue

danny007in:
Fixed https://github.com/ColorlibHQ/AdminLTE/issues/3556


zammitjohn (commented):
Do not use unescape to decode URIs, use decodeURI instead

zammitjohn (approved):

zammitjohn (approved):

PR #3580, opened 2021-04-09
Fixed dropdown-item overlapping dropdown-menu border in active and hover state

78wesley:
Fixed dropdown-item overlapping dropdown-menu border in active and hover state

Dropdown-item overlapped the dropdown-menu when the state active and hover are used on the dropdown-item.
the border-radius is added for the first and last dropdown-item so the round edges looks awesome.

**From:**
Hover state & Active state
image

**To:**
Hover state & Active state
image

### DIFF
--- a/dist/css/adminlte.css
+++ b/dist/css/adminlte.css
@@ -3459,19 +3459,19 @@ input[type="button"].btn-block {
 }
 
 div.dropdown-menu a.dropdown-item:first-child:hover {
-  border-radius: .25rem .25rem 0 0;
+  border-radius: 0.25em 0.25em 0 0;
 }
 
 div.dropdown-menu a.dropdown-item:first-child:active {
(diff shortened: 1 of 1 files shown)


danny007in:
do in scss

`npm install`
and
`npm run dev`

run in command prompt

once check https://github.com/ColorlibHQ/AdminLTE/blob/master/.github/CONTRIBUTING.md

also check some PR(Pull Request)

REJack:
That's already builtin with an extra class.

> To get a bigger dropdown menu you can add .dropdown-menu-lg or .dropdown-menu-xl to .dropdown-menu. You can also get the legacy dropdown offset with adding .dropdown-legacy to .main-header.

REJack:
Oh I missread 🤣

danny007in:
cleared in https://github.com/ColorlibHQ/AdminLTE/issues/3581

PR #3594, opened 2021-04-13
added initial files setup

lhaynes96:


danny007in:
add only initial files not dist

PR #3635, opened 2021-04-26
upd

axelcuaxilo:


danny007in:
i think so ur beginner
check https://github.com/ColorlibHQ/AdminLTE/blob/master/.github/CONTRIBUTING.md

PR #3694, opened 2021-05-15
Fix overflow style on info-box-content class.

Shidersz:
Currently, if the tittle or description of an `info-box` widget is too long, there is an overflow issue, as shown on next image:

info-box-overflow

This PR fix that issue adding the `overflow:hidden` property to the `info-box-content` class to solve that issue, as shown on next image:

info-box-no-overflow

However, note that ellipsis are used when text overflows, mainly because inner elements have the `text-overflow: ellipsis;`  and `white-space: nowrap;` properties, as shown next:

```css
.progress-description,
.info-box-text {
    display: block;
    overflow: hidden;
    text-overflow: ellipsis;
    white-space: nowrap;
}
```

We can also remove those properties, to get a long description or title spawn over multiple lines, but I'm not sure if that one is the desired effect. Check next image for an example:

info-box-wrap

### DIFF
--- a/build/scss/_info-box.scss
+++ b/build/scss/_info-box.scss
@@ -48,6 +48,7 @@
     line-height: 1.8;
     flex: 1;
     padding: 0 10px;
+    overflow: hidden;
   }
 
   .info-box-number {


danny007in:
just add `text-truncate` in scss

Shidersz:
@danny007in Add `text-truncate` to what element, can you give some details?

REJack:
I like the fix, also `.progress-description` & `.info-box-text` have already the style like `text-truncate` has. `text-truncate` doesn't do the trick, @Shidersz found the correct fix for this.

Issue #3716, opened 2021-05-22
[BUG] Control Sidebar do not work when using data-controlsidebar-slide="false"


Shidersz:
**Describe the bug**
The control sidebar do not show when using `data-controlsidebar-slide="false"` as documented [here](https://adminlte.io/docs/3.1//javascript/control-sidebar.html)

**To Reproduce**
1. On a local environment, edit `index.html` demo example.
2. Then replace the control sidebar toggle button definition by next one.
    ```html
    <li class="nav-item">
      <a class="nav-link" data-widget="control-sidebar" data-controlsidebar-slide="false" href="#" role="button">
         <i class="fas fa-th-large"></i>
       </a>
    </li>
    ```

**Expected behavior**
The control sidebar should overlay the content-wrapper without the slide animation.

**Video**
You can see the issue on the next video.

https://user-images.githubusercontent.com/63609705/119242222-85a6eb80-bb32-11eb-95b2-43f976543da1.mp4

**Environment (please complete the following information):**
- AdminLTE Version: v3.1.0
- Operating System:  Linux Mint 20.1
- Browser (Version): Firefox (Latest)

**Additional context**

On the video you could note the class `.control-sidebar-open` is added to the body tag, but the `display: block` style added by that class to the `.control-sidebar` is filtered out, as shown on next image:

![Screenshot from 2021-05-22 19-09-55](https://user-images.githubusercontent.com/63609705/119242255-cc94e100-bb32-11eb-9ff3-f36250c33361.png)

REJack:
fixed with d2e2987a5fdb36151e0f2c90dd1f9b476e33309b

PR #3717, opened 2021-05-22
Fix control sidebar when not using the slide animation

dfsmania:
This PR includes a fix to the control sidebar when using the `data-controlsidebar-slide="false"` option. Related to issue #3716 

You can see it working on next video:

https://user-images.githubusercontent.com/63609705/119242705-98232400-bb36-11eb-92e4-d7df1fda8eea.mp4

Also `index3.html` demo page was modified to include that option, so it is easy to test on the future.

### Alternative Solution.

Instead of changing the `ControlSidebar.js`, we can change style of `.control-sidebar-open` class in `_control-sidebar.scss` to add `display: block !important` instead of just `display: block`, as shown next:

```scss
// Control sidebar open state
.control-sidebar-open {
  .control-sidebar {
    display: block !important;   /* HERE */

    &,
    &::before {
      right: 0;
    }
  }

  &.control-sidebar-push,
  &.control-sidebar-push-slide {
    .content-wrapper,
    .main-footer {
      margin-right: $control-sidebar-width;
    }
  }
}
```

So, let me know what solution is better...

### DIFF
--- a/build/js/ControlSidebar.js
+++ b/build/js/ControlSidebar.js
@@ -97,6 +97,7 @@ class ControlSidebar {
       })
     } else {
       $body.addClass(CLASS_NAME_CONTROL_SIDEBAR_OPEN)
+      $(this._config.target).show()
     }
 
     this._fixHeight()
--- a/index3.html
+++ b/index3.html
@@ -155,7 +155,7 @@
         </a>
       </li>
       <li class="nav-item">
-        <a class="nav-link" data-widget="control-sidebar" data-slide="true" href="#" role="button">
+        <a class="nav-link" data-widget="control-sidebar" data-controlsidebar-slide="false" href="#" role="button">
           <i class="fas fa-th-large"></i>
         </a>
       </li>


danny007in (approved):

Shidersz:
@danny007in I'm going to replace the old `data-slide="true"` option by `data-controlsidebar-slide="true"` on all the demo pages. Is that ok?

danny007in:
> 
> 
> @danny007in I'm going to replace the old `data-slide="true"` option by `data-controlsidebar-slide="true"` on all the demo pages. Is that ok?

No
 it will become braking change

Shidersz:
But `data-slide="true"` won't have any effect now. The `ControlSidebar.js` plugin only read the option `data-controlsidebar-slide`. What's the sense to keep old usage on all the examples?

Shidersz:
> @danny007in I'm going to replace the old `data-slide="true"` option by `data-controlsidebar-slide="true"` on all the demo pages. Is that ok?

I have these changes ready. @REJack please tell me if these replacements are needed...

PR #3803, opened 2021-06-22
Update .gitattributes to trim php package size

lakuapik:
first of all, thanks for the awesome package!

currently, when we run `composer require almasaeed2010/adminlte` we get HUGE package size, its about ~107MB.
one solution to reduce the size of php package is to add `.gitattributes` file, see https://php.watch/articles/composer-gitattributes.

this PR update the `.gitattributes` file to save approximately ~27MB of disk space.

feel free to add/remove entry from `.gitattributes` file as i dont really know what is really needed or not in the php package.
but i am sure the current entry i put is not needed.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,22 @@
 # Ignores for analysis is used to produce the language stats bar which displays the languages percentages
 plugins/* linguist-vendored=true
 docs/* linguist-vendored=true
+
+# Add export-ignore to trim php packagist download size
+# See https://php.watch/articles/composer-gitattributes
+.github/                        export-ignore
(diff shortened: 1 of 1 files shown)


REJack:
The problem is that this change would cut the files out even on the source code download via github releases.

danny007in:
@lakuapik do like .npmignore if possible in composer

lakuapik:
@REJack @danny007in thanks for pointing it out. ill try to find  another alternatives if possible.

REJack:
@lakuapik You can exclude files via a plugin for composer.

Issue #3872, opened 2021-08-02
[BUG] IFrame default tab height issue


Shidersz:
**Describe the bug**
The [IFrame documentation](https://adminlte.io/docs/3.1//javascript/iframe.html) says we have to explicitly define the `height` of a default tab as shown next:

```html
<div class="tab-pane fade active show" id="panel-index" role="tabpanel" aria-labelledby="tab-index">
  <iframe src="./index.html" style="height: 671px;"></iframe>
</div>
```

That is not correct, because height always depend on the client screen size. Height of the default tab should be automatically calculated as with the **empty tab** or the **loading screen**.

**To Reproduce**
1. Install a development environment following the [crontributing guidelines](https://github.com/ColorlibHQ/AdminLTE/blob/a9094a818923204ca4f9d8ed3b2bfb60e615b8df/.github/CONTRIBUTING.md)
2. Edit the `iframe.html` file and use next `content-wrapper` markup to add a default tab
    ```html
    <!-- Content Wrapper. Contains page content -->
    <div class="content-wrapper iframe-mode" data-widget="iframe" data-loading-screen="750">
      <div class="nav navbar navbar-expand navbar-white navbar-light border-bottom p-0">
        <div class="nav-item dropdown">
          <a class="nav-link bg-danger dropdown-toggle" data-toggle="dropdown" href="#" role="button" aria-haspopup="true" aria-expanded="false">Close</a>
          <div class="dropdown-menu mt-0">
            <a class="dropdown-item" href="#" data-widget="iframe-close" data-type="all">Close All</a>
            <a class="dropdown-item" href="#" data-widget="iframe-close" data-type="all-other">Close All Other</a>
          </div>
        </div>
        <a class="nav-link bg-light" href="#" data-widget="iframe-scrollleft"><i class="fas fa-angle-double-left"></i></a>
        <ul class="navbar-nav overflow-hidden" role="tablist">
          <li class="nav-item active" role="presentation">
            <a class="nav-link active" data-toggle="row" id="tab-index" href="#panel-index" role="tab" aria-controls="panel-index" aria-selected="true">
              Dashboard v1
            </a>
          </li>
        </ul>
        <a class="nav-link bg-light" href="#" data-widget="iframe-scrollright"><i class="fas fa-angle-double-right"></i></a>
        <a class="nav-link bg-light" href="#" data-widget="iframe-fullscreen"><i class="fas fa-expand"></i></a>
      </div>
      <div class="tab-content">
[...]

Issue #3822, opened 2021-07-05
[BUG] Content height bugs when page loads


arthurmgb:
I installed AdminLTE on Laravel and I always come across this bug. What happens is that when I load the page for some reason the content height is calculated wrongly and this white "div" in the footer appears for a few seconds, I tried many solutions on the internet and none of them solved my problem 100%. I also tried modifying the .wrapper, .content-wrapper classes, but without success. I also tested it downloading from this repository and the same thing happens. How do I solve this problem?

![124405437-07368f80-dd15-11eb-8677-364442713ada](https://user-images.githubusercontent.com/62814691/124491713-f4aa6d80-dd89-11eb-8d3a-2e249bc097d7.png)

Shidersz:
I want to share some extra data related to this problem:

- There should not be a footer defined on the page.
- To reproduce the problem you can edit the [starter.html](https://github.com/ColorlibHQ/AdminLTE/blob/master/starter.html) page to remove the **footer** definition on a development environment.

arthurmgb:
I deleted the footer and made some changes. I'm not sure if it's right, but this code is working for me:

`            .wrapper{
                width: 100%;
                min-width: 100%;
                height: auto !important;
                position: absolute;
            }
            .content-wrapper{
                position: fixed;
                width: auto;
                top: 50px !important;
                bottom: 0px !important;
                right: 0px !important;
                left: 0px !important;
            } `

REJack:
fixed with 4a2a28f

Shidersz:
@REJack I'm not sure if problem is fixed now. The height was calculated fine previously, but there was a few hundred milliseconds where you see a white rectangle at the bottom of the page (as shown on the image). I have download latest commit to my test environment and the effect is still present, even more, now I can see a vertical scroll bar on the starter page when you do not need one for it. Check next video (look at the bottom)...

https://user-images.githubusercontent.com/63609705/125381611-9a517400-e36a-11eb-90e6-cbcccefd62f9.mp4

I think problem is related with delay on calling method `fixLayoutHeight()`. I will try to investigate more when get some time...

PR #3876, opened 2021-08-07
Fix height problem on default tabs for the IFrame plugin

Shidersz:
In relation to #3822

- Revert some changes that where introduced on the fix https://github.com/ColorlibHQ/AdminLTE/commit/4a2a28f799146867c6b0717936cb14de082e9792 because they do not solve the intended issue #3822 and introduced some other problems related to the height.

In relation to #3872

- Add a fix to solve height calculation of a default tab on the **IFrame** plugin.
- Change documentation of the **IFrame** plugin, to remove hardcoded height for the default tabs. Height can't be hardcoded because height depends on the client screen.

### DIFF
--- a/build/js/IFrame.js
+++ b/build/js/IFrame.js
@@ -227,14 +227,16 @@ class IFrame {
   // Private
 
   _init() {
-    if ($(SELECTOR_TAB_CONTENT).children().length > 2) {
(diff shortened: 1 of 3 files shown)


FlyArtist:
the file: `build/js/IFrame.js` , how should I use the change by `adminlte.min.js?v3.1`,  when I unset the height of iframe, but it's still not fill the screen.

PR #3891, opened 2021-08-14
fix capitalization for overlayscrollbars

tacman:
Otherwise, the library won't load from vendor on linux.


danny007in:
thanks @tacman

Issue #3649, opened 2021-05-02
[BUG] Sidebar links not working on dev-master environment


Shidersz:
**Describe the bug**
After installing the development environment following the steps from [contributing guidelines](https://github.com/ColorlibHQ/AdminLTE/blob/master/.github/CONTRIBUTING.md) I have noted that the links (anchor tags `<a>`) from the sidebar menu are not working for the `click` event.

**To Reproduce**
Steps to reproduce the behaviour:
1. Follow [contributing guidelines](https://github.com/ColorlibHQ/AdminLTE/blob/master/.github/CONTRIBUTING.md) to install a local development environment
2. Run the command `npm run dev`. This will open a live demo at `localhost:3000`
3. Access the live demo using a browser and click on sidebar links.
4. See error (nothing happen, links don't open).

**Expected behavior**
The sidebar links should work for the `click` event.

**Environment (please complete the following information):**
- AdminLTE Version: **dev-master**
- Operating System:  Linux Mint 20.1
- Browser (Version): Mozilla Firefox 88.0 (64 bits)

**Additional context**

After digging a little into this problem, I have noted (using inspector tool from the browser) that the next code from **IFrame.js** file is registering an event listener for the `click` event on the sidebar links:

```js
$(document).on('click', `${SELECTOR_SIDEBAR_MENU_ITEM}, ${SELECTOR_SIDEBAR_SEARCH_ITEM}`, e => {
      e.preventDefault()
      this.openTabSidebar(e.target)
})
```

The `e.preventDefault()` call may be the cause of the problem (prevents the links to open). Even more, looking at the history changes for the **IFrame.js** file, we can see that the below commit moves the method `_setupListeners()` from `_initFrameElement()` to `_init()`. And actually, that method registers the previous mentioned `click` event listener.

https://github.com/ColorlibHQ/AdminLTE/commit/5b07bdd8e9c6710f0b5208dfedf265c418af5b57

As a side note, if I comment the `_setupListeners()` line from the `_init()` method the links works again.

Shidersz:
Note this may be fixed with next commit:

https://github.com/ColorlibHQ/AdminLTE/commit/1f1d19ad528cbed90dd19b336431471111d60f72

REJack:
Yep that was inteded to fix it 😄 I only forgot to write it here too 🤦🏻‍♂️

Shidersz:
[...]

PR #3897, opened 2021-08-16
Fix logic to capture links on IFrame mode.

Shidersz:
Minor fix on the logic that capture links on the **IFrame** mode. Related to issue #3649


REJack:
🤦🏻‍♂️ I fixed it by my self before 20mins or so, I didn't registered this issue :D so I close it :)

PR #3990, opened 2021-10-13
Improve ExpandableTable

hkvstore:
1. Add expand/collapse events so users can abort expanding/collapsing by ``event.preventDefault()``.
2. Add ``relatedTarget`` to expanded/collapsed events. The related target tells the actual HTML element clicked in the row so users can determine what to do in the "expanded" (or "collapsed") event (e.g. show different content in the expanded row).
3. Simplify initialization by CSS.


hkvstore:
Withdraw

Issue #4003, opened 2021-10-24
[BUG] Pagination border color on dark mode (Yes, that small sh**)


SmileYzn:
**Describe the bug**
Fix the border color for pagination buttons when using on dark mode

**To Reproduce**
1. Go to dark mode
2. go to pagination component
3. see the border color is not changed yet

**Expected behavior**
The normal dark color replacement

**Screenshots**
https://i.imgur.com/BRdENf9.png

**Environment (please complete the following information):**
- AdminLTE Version: 3.10 (Via jsdelivr)
- Operating System:  Solux linux
- Browser (Version): Firefox latest

**Additional context**
Check the rule:
`.page-item.active .page-link {
	z-index: 3;
	color: #fff;
	background-color: #007bff;
	border-color: #007bff;
}`

at _pagiantion.scss

:P

PR #4004, opened 2021-10-24
Fixed border color for active item on dark-mode

SmileYzn:
Fixed border color for active item on dark-mode


SmileYzn:
Issue #4003

PR #4091, opened 2021-11-28
feature

laraveladmin-cn:
Support Vue single page application content is not "body"
支持Vue单页面应用内容非”body”的情况


REJack:
We actually don't accept PR's for AdminLTE v4 Plugins, since there are not portet and get a complete overhaul.

PR #4131, opened 2021-12-13
Added remember state option to pushmenu/sidebar menu

gintsmurans:
As title says, this will add option to remember menu via data attributes.

### DIFF
--- a/src/ts/push-menu.ts
+++ b/src/ts/push-menu.ts
@@ -14,6 +14,17 @@ import {
  * Constants
  * ------------------------------------------------------------------------
  */
+const DATA_KEY = 'lte.pushmenu'
+const EVENT_KEY = `.${DATA_KEY}`
+
+const EVENT_EXPAND = `expand${EVENT_KEY}`
+const EVENT_COLLAPSE = `collapse${EVENT_KEY}`
+const EVENT_CLOSE = `close${EVENT_KEY}`
+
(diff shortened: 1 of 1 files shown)


REJack:
Overall a nice PR, but I'm personally not a friend of cookies, I would preferr a option between cookie & localStorage and also I miss the enable/disable option for it.

gintsmurans:
Those are some good points, updated code.

lenamtl:
Is this possible to have the JS compiled file for this for a temporary solution ?

lenamtl:
Hi,
I'm having issue with this code, 
This occurred when url have a # fr example **index4.php#**
Then refresh the page 
all items are now selected 

To reproduce
let say we are on index.php
we click on vertical menu parent item to open sub-menu the parent item has `<a href="#" class="nav-link ">`
so now the url is index.php#
refresh the page
all menu item are now active

danny007in:
add true or false in toRemeber in _config

danny007in:
can you add here https://github.com/ColorlibHQ/AdminLTE/blob/v4-dev/src/partials/_scripts.html

danny007in:
fixed in https://github.com/ColorlibHQ/AdminLTE/commit/93c1e370fa40fde73e34e53285afed2a2cc4a0db

danny007in:
Make PR commit readable https://www.youtube.com/watch?v=gXCkYkLQ3To

Issue #4255, opened 2022-02-08
[BUG] Top-level selectors may not contain the parent selector "&"


forxer:
**Error on local NPM build with v3.2.0**

```
ERROR in ./resources/sass/admin/app.scss
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: Top-level selectors may not contain the parent selector "&".
  ╷
7 │   &.bg-#{$name} {
  │   ^^^^^^^^^^^^^^
  ╵
  node_modules\admin-lte\build\scss\mixins\_backgrounds.scss 7:3  background-variant()
  node_modules\admin-lte\build\scss\_colors.scss 7:3              @import
  resources\vendor\adminlte3\sass\parts\_miscellaneous.scss 9:9   @import
  resources\vendor\adminlte3\sass\adminlte.scss 25:9              @import
```

**Environment:**
- AdminLTE Version: **v3.2.0**
- Operating System:  Windows 10

**Additional informations**
I've fix this error by replacing, in build\scss\mixins\_backgrounds.scss line 7:

`&.bg-#{$name} {`

by

`.bg-#{$name} {`

There may be other similar occurrences but I am not including all files in my project.

forxer:
Please, release hotfixes more regularly. 

"Release early, release often" :)

glavrjk:
Same problem.

admin-lte 3.2.0
Win10
sass 1.49.7
sass-loader 12.4.0
@symfony/webpack-encore 1.8.1

ChengXuchao:
您好，您的来信我已经收到！我会及时与您取得联系。谢谢……

picks44:
Same issue here.

admin-lte 3.2.0
Win10
sass 1.49.7
sass-loader 12.4.0
@symfony/webpack-encore 1.8.1

REJack:
I search actually for a better solution, if I revert the change some overwrites doesn't work anymore.

mbeckerle-xqueue:
I have the same issue. In fact, downgrading to an older version or manually removing the & in _background.scss works but I also have no idea how to resolve this properly.

In this comment you say "bg-* get overwritten by print mode"
https://github.com/ColorlibHQ/AdminLTE/commit/52d947384e057378fc65b5b5f421caa9318be66a
May I ask: What exactly was the problem?

REJack:
[...]

PR #4276, opened 2022-02-19
Add verification use of & outside of style rules

apteles:
This solves the error when building in Laravel mix. On other hand, this error was already mentioned in #4255. 

See the error:

```bash
SassError: Top-level selectors may not contain the parent selector "&".
   ╷ 
7 │   &.bg-#{$name} {
   │   ^^^^^^^^^^^^^^
   ╵
```

How it was solved:

See documentation: https://sass-lang.com/documentation/style-rules/parent-selector#in-sassscript where says:

`If the & expression is used outside any style rules, it returns null. Since null is [falsey](https://sass-lang.com/documentation/at-rules/control/if#truthiness-and-falsiness), this means you can easily use it to determine whether a mixin is being called in a style rule or not`

I hope that it helps other people.

### DIFF
--- a/build/scss/mixins/_backgrounds.scss
+++ b/build/scss/mixins/_backgrounds.scss
@@ -4,7 +4,7 @@
 
 // Background Variant
 @mixin background-variant($name, $color) {
-  &.bg-#{$name} {
+  #{if(&, "&.bg-#{$name}",".bg-#{$name}")} {
     background-color: #{$color} !important;
 
     &,


alexandre-castelain:
With this fix, I can build my assets. Thanks

nathabonfim59:
As of now, this fix is not included in the release of version 3 and is unlikely to be added, especially since this is an older version and the focus is on the new one, which is absolutely fair.

If you're stuck using an older version like me, you don't need to wait for a new release, simply add the following line to your package.json:

```json
    "admin-lte": "ColorlibHQ/AdminLTE#v3",
```

> [!Warning]
> Please be aware that this change can cause compatibility issues or unexpected behavior, so testing in a development environment is recommended before deploying to production. The `*.lock` files are your best allies here.

PR #4456, opened 2022-05-14
Fix: Customize AdminLTE

darkterminal:
Hi, I Have fixed Customize AdminLTE in `demo.js`

**Added**
- `localStorage` to save Preference
- Load preferences


REJack:
1. nice work but we doesn't want a feature like this, otherwise anybody asks about changing it and its only for demo purposes
2. never publish plugin files and dist files with your PR

Issue #4479, opened 2022-05-28
[BUG] Can't maximized a card if it was sorted by using sortable.js


jjsty1e:
**Describe the bug**
If there are two card in a page, and they can be sorted by using jquery-ui/sortable.js, then they would unable to be maximized by click the maximized button(if you have added this button)

**To Reproduce**
1. Let's produce it at the dashboard v1 page
2. add maximize buttons to left card "Sales", and right card "Visitors", the code world be 
```html
<button class="btn btn-tool"
  data-card-widget="maximize">
 <i class="fas fa-expand"></i></button>
```
3. drag Visitors Card to replace Sales Card, click the maximize button on Visitors Card
5. See error: Visitors Card can't be maximized

**Expected behavior**
Visitors Card can be maximized

**Screenshots**
none

**Environment (please complete the following information):**
- AdminLTE Version: [e.g. v3.2.0]
- Operating System:  [e.g. macOS Catalina]
- Browser (Version): [e.g. Chrome (Latest)]

**Additional context**
I found it would be the sortable.js's fault, but sortable.js is a built-in plugin, so I suggest to fix it by changing our's codes, right?

I have found the solution, I will create a PR for this.

REJack:
I've added your fix with 75deb497b39c1a8547fe0f21a2478b066b223396.

jjsty1e:
OK then👍

PR #4480, opened 2022-05-28
Fix CardWidget unable to be maximized after it was sorted.

jjsty1e:
The related issue:  #4479


jjsty1e:
this would lead to a side effect, ie, at the click moment, the card below would go up (although it would go back if the card was minimized)

REJack:
Close due it's already implemented 😉

Issue #4339, opened 2022-03-14
[BUG] the Treeview won't collapse after page refresh with vuejs


galihsetyo:
Hi guys, I run Adminlte with Laravel using Inertiajs with Vue (SPA).
the app is loading the adminlte on the login page so after logging in and being redirected to the dashboard the treeview needs to be initiated. 
I initiate the treeview with this:

`$('[data-widget="treeview"]').each(function () {
                adminlte.Treeview._jQueryInterface.call($(this), 'init')
            })`

and then the treeview is working properly.

but, the problem is when I refresh the page on the dashboard, the treeview can open but won't collapse.

I think it's because the treeview init runs more than once, first when the page refreshed and then when my code initiates the treeview again.
how to check whether the treeview was initiated or not?, so I can condition my code to only run when the treeview is not initiated yet, or is there any other solution with this?, sorry for my bad English, I appreciate the help.

airtonctj:
Hello good Morning,

I have the same problem

=(

tiagocarvalhomk:
No solution for that?

wchar-t:
It does happen because it's being ran twice.
I had to hack a little, in order to make it work.

I noticed that this portion isn't running sometimes - i guess that's because the menu isn't rendered when this setup runs (i'm using React)
![image](https://user-images.githubusercontent.com/8273763/167049150-dee4b31e-5004-4e50-92a0-932272d23c66.png)

But you can avoid that, and init by yourself whenever you want (just make sure you are checking everytime wether or not you have done a init before)
![image](https://user-images.githubusercontent.com/8273763/167049469-4b6592f5-80ae-4f85-9b55-25df2083c6c3.png)

galihsetyo:
Hi, thanks for the reply guys.
Finally, I got it to work by doing this, first I make a variable to check if the treeview has been loaded with the load.lte.treeview event
and then when document.readyState is complete I check for the variable whether it is still undefined or not, only if it is still undefined then I initiate the treeview

```
let treeView;

$(window).on('load.lte.treeview', function (e) {
    treeView = e;
});

var readyStateCheckInterval = setInterval(function() {
    if (document.readyState === "complete") {
        clearInterval(readyStateCheckInterval);

        if(treeView == undefined) {
            $('[data-widget="treeview"]').each(function() {
                adminlte.Treeview._jQueryInterface.call($(this), 'init');
            })
        }
    }
}, 10);
```

galihsetyo:
[...]

PR #4521, opened 2022-06-25
fix: prevent multiple event listener calls #4339

galihsetyo:
in some implementation cases in SPA, the Treeview must be initiated manually, but when the page is refreshed then the Treeview will be initialized more than once and create duplicate event listeners that make the Treeview toggle become buggy (it won't collapse when the Treeview initiated more than once), so I added the stopImmediatePropagation to prevent the multiple event listener calls.


danny007in:
Give some more details regarding to issue like code, images, video to solve issue

PR #4731, opened 2022-10-05
Don't use sidebar-mini in mobile mode

jcheng5:
In mobile mode, the mini sidebar doesn't work well, since it overlaps with the header (it effectively can't be dismissed). 

So instead of going mini, just close.

### Before

https://user-images.githubusercontent.com/129551/193960752-f284b565-8d47-41c1-a2f4-41f91366452e.mov

### After

https://user-images.githubusercontent.com/129551/193960910-d62cf159-8c0c-4356-9880-ecac136090e8.mov

### DIFF
--- a/src/ts/push-menu.ts
+++ b/src/ts/push-menu.ts
@@ -172,6 +172,13 @@ class PushMenu {
   }
 
   toggleMini(): void {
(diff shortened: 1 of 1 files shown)


danny007in:
fixed in https://github.com/ColorlibHQ/AdminLTE/commit/93c1e370fa40fde73e34e53285afed2a2cc4a0db

PR #4750, opened 2022-10-15
fix: published package

THETCR:
Currently the NPM package contains all project files, even the .idea configuration folder for PhpStorm.
The end user that imports your node module doesn't need those files. 

The same thing applies to the documentation, this shouldn't be a part of the published package.
Next to installing those in the dependency tree of the end user, the postinstall lifecycle script also runs for production installations, making an extra copy of the dependencies needed by the documentation in the plugins directory.

The NPM package currently weighs about ~215MB.
While the actual distribution build combined with the build directory and jQuery takes about ~28,5MB.

That is approximately ~186,5MB just to run the documentation.

Most end users won't be using any of it.

Besides the ~56MB of the copied modules in the plugins directory, the unnecessary dependencies cost ~130MB.
Which is always installed, even in CI/CD etc.

This PR changes the following:

- Defines the files to be included in the package through the 'files' field in the package manifest.
- Changes the postinstall script to a Pre lifecycle script for the 'docs' script.
- Moves all dependencies that are not required by the distribution build to the devDependencies.


danny007in:
@REJack Check this PR

THETCR:
@REJack 
It would've been decent to give a reply.

danny007in:
> @REJack It would've been decent to give a reply.

Sorry for inconvenience, AdminLTE v3 is no longer maintained, please shift to v4.

THETCR:
@danny007in 

This could've been a patch version, significantly reducing the resources used.
You know, with the environment and all that.

PR #4626, opened 2022-07-28
fixed typos in `adminlte.css`

KunalTanwar:
this fixes #4623

PR #4916, opened 2023-02-15
Fix incompatibility when calculating border radius

louman:
This addresses the issue initially reported in https://github.com/ColorlibHQ/AdminLTE/pull/4626

I followed the contribution guide but didn't commit the last step of running `npm run production` since it modifies hundreds of files. Please let me know if anything else is needed.

### DIFF
--- a/build/scss/_buttons.scss
+++ b/build/scss/_buttons.scss
@@ -10,7 +10,7 @@
 
   // Flat buttons
   &.btn-flat {
-    @include border-radius(0);
+    @include border-radius(0rem);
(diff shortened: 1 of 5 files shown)


louman:
@REJack would you be able to assist me with fixing the lint issues?

louman:
Seems related to https://stylelint.io/user-guide/rules/length-zero-no-unit/ which prevent units for zero values problem is that `sass` won't compile saying `Error: 0.25rem and 0 are incompatible.`

PR #5084, opened 2023-04-25
Replace node-sass with dart-sass

bmorrall:
I was working on a project where I plan to use AdminLTE's theme, but I was having a lot of problems with the sass files being unable to be complied.

From a bit of digging around, I found that I was using `dart-sass` instead of `node-sass` and the way the imports were setup was not compatible with each other.

From further digging, I found that [node-sass is deprecated](https://sass-lang.com/blog/libsass-is-deprecated) so I made this PR to swap over the libraries.

Please let me know if I need to change anything.


danny007in:
https://github.com/ColorlibHQ/AdminLTE/tree/v4-dev contribute here 👍🏼

danny007in:
dart-sass will be breaking change for AdminLTE v3, so dart-sass used in v4

Issue #5296, opened 2023-08-23
[FEATURE] Remove external sources


faca5:
Hello.

Remove all external sources and host local.

Example:
Google Fonts

`<link rel="stylesheet" href="[https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,400i,700&display=fallback](view-source:https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,400i,700&display=fallback)">`

Reason:
GDPR compliance (EU law), Logging IP address from external companies outside EU

Note:
jQuery, Bootstrap etc... already hosted locally.

puikinsh:
Closing — AdminLTE v4 uses npm packages and local assets. External CDN sources are not used in the core template.

PR #5313, opened 2023-09-01
Local Google Fonts

glewe:
This was a feature request #5296
I also ran into cases where I had to switch to local Google fonts due to European GDPR laws (loading fonts from Google transfers the user's IP address to Google without his consent).

The currently used font Source Sans Pro has been updated and is now named "Source Sans 3". I downloaded the woff2 files for weights 300, 400 and 700 (normal and italic) in cyrillic, cyrillic-ext, greek, greek-ext, latin, latin-ext and vietnamese versions (as the remote CSS from Google fonts currently provides).

I left the reference to the remote Google fonts in `_head.astro` in case someone wants to use those. However, in that case the user also needs to remove the references to the local font files and re-compile adminlte.css.

### DIFF
--- a/src/html/components/_head.astro
+++ b/src/html/components/_head.astro
@@ -19,11 +19,11 @@ const cssPath = isRtl ? ".rtl" : "";
 />
 <!--end::Primary Meta Tags-->
 <!--begin::Fonts-->
-<link rel="preconnect" href="https://fonts.googleapis.com" />
-<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
 <link
-  href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,300;0,400;0,700;1,400&display=swap"
   rel="stylesheet"
+  href="https://cdn.jsdelivr.net/npm/@fontsource/source-sans-3@5.0.12/index.css"
+  integrity="sha256-tXJfXfp6Ewt1ilPzLDtQnJV4hclT9XuaZUKyUvmyr+Q="
+  crossorigin="anonymous"
 />
 <!--end::Fonts-->
 <!--begin::Third Party Plugin(OverlayScrollbars)-->
--- a/src/scss/_bootstrap-variables.scss
+++ b/src/scss/_bootstrap-variables.scss
@@ -607,7 +607,7 @@ $aspect-ratios: (
 
 // scss-docs-start font-variables
 // stylelint-disable value-keyword-case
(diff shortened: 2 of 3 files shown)


danny007in:
is there any cdn option?

glewe:
> 

Any option with accessing resources from a remote source before the user gives his consent is legally not allowed. The reason behind this is that a remote service can read and use the requesting IP address which is part of the user's identity. In the case of Google fonts there were quite a lot of law suites last year in Europe, specifically in Germany. This lead for example to the development of the OMGF WordPress plugin that allows to host the used Google fonts locally.

danny007in:
Can we use this

https://www.jsdelivr.com/package/npm/source-sans-pro

glewe:
> After running the `npm run production` command, will this work? I would prefer to use only CDNs so that users can choose their own fonts.

Indeed, `npm run production` does not work with the node_modules version (npm run dev does). It cannot find the .css file so it says. 
So I switched to the CDN of Source Sans Pro as you suggested. I assume that is acceptable from a GDPR perspective since I have never heard of that case being a problem.
I also see your point of making it easier for users to select a different font.

src/assets/fonts/SourceSans3/SourceSans3-Light-cyrillic-ext.woff2
danny007in, line 1: remove unwanted files

src/html/components/_head.astro
```diff
@@ -19,12 +19,10 @@ const cssPath = isRtl ? ".rtl" : "";
 />
 <!--end::Primary Meta Tags-->
 <!--begin::Fonts-->
-<link rel="preconnect" href="https://fonts.googleapis.com" />
-<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
 <link
-  href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,300;0,400;0,700;1,400&display=swap"
   rel="stylesheet"
-/>
+  href=" https://cdn.jsdelivr.net/npm/source-sans-pro@3.6.0/source-sans-pro.min.css "
```
danny007in, line 24: Please visit https://www.jsdelivr.com/package/npm/@fontsource/source-sans-3?tab=files

image

danny007in (changes requested):

src/assets/fonts/SourceSans3/SourceSans3-Light-cyrillic-ext.woff2
glewe: Oops, my bad. Now removed.

src/html/components/_head.astro
glewe: Yes, that is another option. I switched to Source Sans 3 from fontsource.

danny007in:
I'm going to make some changes to the code. Please don't push anything

PR #5344, opened 2023-09-20
theme customizer

danny007in:


danny007in:
@jameslao pushed new code, you can make changes

jameslao:
Hmm I somehow feel the benefit of the original control sidebar is that you can preview the changes on any page. The theme generator works well on its own but can't do that part...?

PR #5536, opened 2024-04-01
implement control sidebar

jameslao:


danny007in:
where is toggle button

jameslao:
> where is toggle button

Just implemented. OverlayScrollbar may need some tweak

danny007in:
https://github.com/ColorlibHQ/AdminLTE/pull/5344 please check
I think theme generator is better that control sidebar

jameslao:
> #5344 please check I think theme generator is better that control sidebar

Absolutely. I'll look at it

jameslao:
> > #5344 please check I think theme generator is better that control sidebar
> 
> Absolutely. I'll look at it

I'd agree. Theme generator can totally serve the purpose. Any specific help you might need before you merge?

danny007in:
@jameslao i will push code and inform you please wait

src/html/components/_scripts.astro
```diff
@@ -4,7 +4,7 @@ const adminlteJsUrl = path + "/js/adminlte.js";
 ---
 
 <!--begin::Third Party Plugin(OverlayScrollbars)-->
```
Navisystem77, line 6:
```suggestion
<!--begin::Third Party Plugin(OverlayScrollbars)-->
```

puikinsh:
Thank you for the effort on this @jameslao! After reviewing the discussion here and considering the direction of v4, we've decided to go with the theme generator page approach instead of a control sidebar panel.

As you and @danny007in discussed, the theme generator provides a cleaner solution that fits better with v4's architecture. The control sidebar pattern from v3 relied on conventions (`data-widget`) that we've moved away from in v4.

We appreciate your contribution and the time you put into this! 🙏

PR #5715, opened 2024-09-27
scss fix for build

ZaidPathan:
&.bg-#{$name} { -> #{if(&, '&.bg-#{$name}','.bg-#{$name}')} {

### DIFF
--- a/build/scss/mixins/_toasts.scss
+++ b/build/scss/mixins/_toasts.scss
@@ -4,10 +4,9 @@
 
 // Toast Variant
 @mixin toast-variant($name, $color) {
-  &.bg-#{$name} {
-    background-color: rgba($color, .9) !important;
+  #{if(&, '&.bg-#{$name}','.bg-#{$name}')} {
+    background-color: rgba($color, 0.9) !important;
     @if (color-yiq($color) == $yiq-text-light) {
-
       .close {
         color: color-yiq($color);
         text-shadow: 0 1px 0 $black;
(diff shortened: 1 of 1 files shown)


danny007in:
For guidance on how to write clear and effective PR commit messages, you can watch this [YouTube video](https://www.youtube.com/watch?v=gXCkYkLQ3To).

build/scss/mixins/_toasts.scss
```diff
@@ -4,20 +4,18 @@
 
 // Toast Variant
 @mixin toast-variant($name, $color) {
-  &.bg-#{$name} {
-    background-color: rgba($color, .9) !important;
+  #{if(&, '&.bg-#{$name}','.bg-#{$name}')} {
+    background-color: rgba($color, 0.9) !important;
```
theneonstark, line 8: Your part attempts to use a conditional to check if the current selector is & and the syntax here is incorrect and might not achieve your intended outcome.

build/scss/mixins/_toasts.scss
```diff
@@ -4,20 +4,18 @@
...
-
       .close {
         color: color-yiq($color);
         text-shadow: 0 1px 0 $black;
       }
     }
 
     .toast-header {
-      background-color: rgba($color, .85);
+      background-color: rgba($color, 0.85);
```
theneonstark, line 17: I think this gonna be right but no matter use .85 or 0.85 because SCSS detect the same for this

theneonstark (changes requested):

PR #5814, opened 2025-01-25
Add storage key for sidebar state

burdittw:
Added the TS needed to store the sidebar state in the browsers local storage. This will store the state of the sidebar menu so that if it is collapsed or expanded, when the page is reloaded it will render in the same state the user wanted.


danny007in:
make save state as option, please do not push dist instead src

puikinsh:
Added proper implementation of this in https://github.com/ColorlibHQ/AdminLTE/commit/e4493360a44440d6c915eb541e1c8f4d2cba5bc5

PR #5928, opened 2025-11-09
Update workflows (github actions)

dfsmania:
Update all workflows to use **node 20** and **setup-node@v4**.
Fix the currently failure Lint action.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,7 +9,7 @@ on:
 
 env:
   FORCE_COLOR: 2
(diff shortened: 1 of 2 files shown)


puikinsh:
@dfsmania Thanks for pointing this out. 
I have pushed things slightly further with the Node 22 for a longer-term support, but still all the credits go to you. 

See: https://github.com/ColorlibHQ/AdminLTE/commit/e3a3ca5d0f96c41d7eb57b7e41884a4f5eecf2c9

PR #5953, opened 2025-12-24
Remove code comments on TypeScript compilation process

dfsmania:
Remove comments on the compilation process of `TypeScript` code. This only affects the generated `adminlte.js` file, as the minimized `adminlte.min.js` already handles this.
- Comments are useful only for maintenance, on source files, there's no benefit on having them on the compiled `adminlte.js` distribution file.
- This also reduces the compiled file sizes.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "root": true,
   "compilerOptions": {
+    "removeComments": true,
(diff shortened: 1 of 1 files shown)


puikinsh:
Thanks @dfsmania! Merged this into master (commit 68e95c10). The unminified `adminlte.js` dropped by ~15% — nice cleanup. 👍

PR #5906, opened 2025-09-04
Fix floating labels with firefox

dfsmania:
### Summary

This PR fixes an issue with **floating labels** in `Firefox` where labels appear in the wrong position due to `:-moz-placeholder` rules being generated in the compiled CSS.

Fix #5901 

#### Browserslist update

```diff
+ not kaios <= 2.5 # fix floating label issues in Firefox (see https://github.com/postcss/autoprefixer/issues/1533)
```
- This prevents **Autoprefixer** from generating obsolete `-moz-placeholder` selectors.
- **Bootstrap** applies the same workaround in their own Browserslist config.
- Without this change, `Firefox` treats inputs as if they always contain content, breaking floating labels.

#### Watch task update

```diff
- "watch-css-main": "nodemon --watch src/scss/ --ext scss --exec \"npm-run-all css-lint css-compile copy-assets\"",
+ "watch-css-main": "nodemon --watch src/scss/ --ext scss --exec \"npm-run-all css-lint css-compile css-prefix copy-assets\"",
```
- Adds `css-prefix` command to `watch-css-main` task.
- This ensures `npm run dev` applies the same `PostCSS` prefixing step as `npm run production`.
- Keeps development and production builds consistent, so issues like the Firefox floating label bug can be detected early.

@danny007in @puikinsh please merge as soon as you can, thanks!

PR #5955, opened 2025-12-26
Update .browserslistrc

dolva2000:


dfsmania:
@dolva2000 Please, explain the benefits of these changes. Actually you worked on an outdated file, because I pushed a fix on this file recently and it's not present here (see #5906 ), so it's having merging conflicts. Try to update your repo to the latest version before creating a PR.

puikinsh:
Closing — our current `.browserslistrc` already targets modern browsers (Chrome 97+, Safari 15.4+, Firefox 104+) which is more appropriate for 2026. This PR would regress to older targets and duplicates the rules multiple times.

PR #5956, opened 2025-12-26
Refactor layout transition handling and initialization logic

dfsmania:
## Refactor layout transition handling and initialization logic

This PR refactors the **AdminLTE v4 layout logic** to fix incorrect transition handling, remove duplicated event listeners, and clearly separate responsibilities between initialization and runtime behavior.

### 1. Removed duplicated `resize` listeners

* The `resize` event listener was previously:

  * registered **inside** `Layout.holdTransition()`
  * invoked from multiple entry points
* This caused:

  * duplicated listeners
  * unpredictable execution order
  * unnecessary re-registration on each call

The resize listener is now registered **once**, during layout initialization.

### 2. Fixed incorrect `holdTransition()` logic

The previous implementation had a critical flaw:

* `resizeTimer` was defined as a **local variable**
* `clearTimeout(resizeTimer)` was ineffective across calls
* Each resize created a new timer reference, defeating its purpose

#### New behavior:

* `Layout` now owns a persistent `_holdTransitionTimer`
* Any pending timer is properly cleared before starting a new one
* Transitions are reliably held and released

This makes `holdTransition()` **deterministic and correct**.

### 3. Clear responsibility for `holdTransition()`

`holdTransition()` now does exactly one thing:

* Temporarily disables layout transitions for a given duration

It no longer:

* registers event listeners
* depends on resize side effects
* mixes setup with behavior

This makes the method reusable and predictable.

### 4. Resize logic moved to initialization phase

* Window resize handling is now defined in the Data API section
* Resize simply calls `layout.holdTransition(200)`
* The layout class no longer manages global events

This keeps lifecycle concerns outside the core class.

### 5. Layout initialization cleaned up in `adminlte.ts`

* Removed duplicated layout initialization logic
* Layout setup now happens in a single place (`layout.ts`)
* Avoids creating multiple layout instances with overlapping responsibilities

### 6. App loaded state handled consistently

* The `app-loaded` class is now applied from the layout initialization logic
* Prevents scattered timing logic across multiple files
* Makes layout readiness easier to reason about

---

### Why this change?

The previous layout implementation:

* mixed event registration and behavior
* contained broken timeout logic
* created multiple resize listeners
* duplicated layout initialization across files

This refactor:

* fixes real bugs
* simplifies the mental model
* improves maintainability
* aligns layout behavior with a clean, instance-based design

No visual or behavioral regressions are expected—only more predictable and correct transition handling.

### DIFF
--- a/src/ts/adminlte.ts
+++ b/src/ts/adminlte.ts
@@ -16,13 +16,6 @@ import { initAccessibility } from './accessibility.js'
 
 onDOMContentLoaded(() => {
   /**
-   * Initialize AdminLTE Core Components
-   * -------------------------------
-   */
-  const layout = new Layout(document.body)
-  layout.holdTransition()
-  
-  /**
    * Initialize Accessibility Features - WCAG 2.1 AA Compliance
    * --------------------------------------------------------
    */
@@ -33,14 +26,9 @@ onDOMContentLoaded(() => {
     keyboardNavigation: true,
     reducedMotion: true
   })
-  
+
   // Add semantic landmarks
   accessibilityManager.addLandmarks()
-  
-  // Mark app as loaded after initialization
-  setTimeout(() => {
-    document.body.classList.add('app-loaded')
-  }, 400)
 })
 
 export {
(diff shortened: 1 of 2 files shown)


puikinsh:
Thanks @dfsmania! Merged into master (commit 26bc0358). The broken timer and duplicate listener issues were great catches — layout transitions are now deterministic. 👍

PR #5963, opened 2026-01-03
Increase width of auth login/register box

dfsmania:
**Summary**
Increase the width of the authentication login/register box from `360px` to `400px`.


puikinsh:
Thanks @dfsmania! Merged into master (commit d4ee248b). Good call on the wider box — 360px was a bit cramped. 👍

PR #5964, opened 2026-01-09
fix: early return from slideDown/slideUp if duration is less than 1

Kneemund:
This PR adds an early return to the `slideDown` / `slideDown` function that skips the animation if the duration is less than (or equal to) 1. 

---

Currently, when the duration is zero, the second timeout runs BEFORE the first one (at least on Firefox), and doesn't properly clear the style attributes:

https://github.com/ColorlibHQ/AdminLTE/blob/5196a37407d9df1d1ac35d13c86e7bd1a5448282/src/ts/util/index.ts#L96-L112

---

This fixes a bug in the sidebar treeview, where this function is used with a duration of zero to instantly open entries when the page loads:

https://github.com/ColorlibHQ/AdminLTE/blob/5196a37407d9df1d1ac35d13c86e7bd1a5448282/src/ts/treeview.ts#L120

As an alternative fix, the duration parameter could be clamped to be greater than 1.

### DIFF
--- a/src/ts/util/index.ts
+++ b/src/ts/util/index.ts
@@ -48,6 +48,11 @@ const safePropertyAccess = (obj: Record<string, unknown>, property: string): unk
 
 /* SLIDE UP */
 const slideUp = (target: HTMLElement, duration = 500) => {
+  if (duration <= 1) {
+    target.style.display = 'none'
+    return
+  }
+
   target.style.transitionProperty = 'height, margin, padding'
   target.style.transitionDuration = `${duration}ms`
   target.style.boxSizing = 'border-box'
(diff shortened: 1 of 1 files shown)


dfsmania:
@Kneemund would this same fix applicable to `slideUp()`?
I'm also trying to improve `Typescript` logic of this package when I got some free time.

Kneemund:
That function is never used with a duration of 0. But you're right, it *could* happen, so the fix should be applied to that function too.

puikinsh:
Thanks @Kneemund! Merged into master (commit 0d4d8577). Great catch on the setTimeout ordering issue — the early return approach is clean and correctly handles the zero-duration case used by treeview. 👍

PR #5999, opened 2026-03-12
feat: add ThemeSwitch component for dark/light mode toggle

aiheroshare-vibe-coder:
## Summary

- Новый компонент ThemeSwitch (src/ts/theme-switch.ts) — 
  переключение тёмной/светлой темы через data-lte-toggle="theme"
- Сохранение выбора в localStorage
- Поддержка системной настройки prefers-color-scheme
- Экспорт из главного adminlte.ts

## Использование

<button data-lte-toggle="theme">Toggle Dark Mode</button>


dfsmania:
Just a warning, this PR introduces unintended changes!

puikinsh:
Thanks for the contribution, @aiheroshare-vibe-coder! The use case here is solid and exactly what users have been asking for in #6010.

Closing this in favour of an alternative approach that just landed: a Light / Dark / Auto dropdown was added directly to the default topbar (`src/html/components/dashboard/_topbar.astro`), with the wiring shipped globally in `src/html/components/_scripts.astro`. This means the toggle is visible and working out of the box on every demo page — no markup setup required from users.

Differences from your proposal:

- Three options (Light / Dark / Auto) instead of a binary toggle, matching the existing `docs/color-mode.html` reference implementation
- `Auto` follows `prefers-color-scheme` and live-updates when the OS theme changes
- Persists to `localStorage` under `lte-theme` (same key your PR used)
- Pure HTML/JS in the topbar — no new TypeScript module to maintain, keeps the JS bundle lean

The end result is the same — users get a working theme toggle out of the box — but it ships without expanding the JS surface. Will be in the next release.

Thanks again for taking the time to put this together. The motivation behind the PR was a real gap and seeing it raised pushed this onto the release.

PR #6003, opened 2026-03-17
fix: consistent sidebar style

Kneemund:
Currently, nested links look different than top level links, which is especially noticeable when using custom sidebar colors. This PR makes the sidebar style more consistent, mostly by expanding the style of nested links onto top level links.

Here are a few of the inconsistencies in the old version fixed by this PR:
- Hovering over active top level links overrides the active colors with the hover colors. This does not happen with nested links.
- Focusing nested links does not apply the hover background color. This does happen with top level links.
- Open menus that are not active look weird, as they have the normal background but the hover text color. Because the SCSS rule explicitly sets the hover background color to transparent, it messes up the focus state (since that reuses the hover colors).

### DIFF
--- a/src/scss/_app-sidebar.scss
+++ b/src/scss/_app-sidebar.scss
@@ -106,31 +106,31 @@
 
   // Sidebar Menu. First level links
   .sidebar-menu > .nav-item {
-    &.menu-open {
-      .nav-link.active:not(:hover) {
-        --#{$lte-prefix}sidebar-menu-active-bg: transparent;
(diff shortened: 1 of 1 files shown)


puikinsh:
Thanks again @Kneemund! Really appreciate you taking the time to untangle the sidebar color specificity issues — the hover/active/focus inconsistencies between top-level and nested links have been bugging us. The simplified selectors are much cleaner. Merging this as well. 🙏

PR #6055, opened 2026-06-21
fix: improve fullscreen handling and accessibility safeguards

webgo-oss:
### Summary

This PR fixes a few JavaScript issues related to fullscreen handling and accessibility utilities.

#### 1. Handle Fullscreen API Promise Correctly

Updated fullscreen toggle logic to wait for `requestFullscreen()` to resolve before changing icon states.

**Before**

```js
document.documentElement.requestFullscreen()
if (iconMaximize) iconMaximize.classList.add('d-none')
if (iconMinimize) iconMinimize.classList.remove('d-none')
```

**After**

```js
document.documentElement.requestFullscreen()
  .then(() => {
    if (iconMaximize) iconMaximize.classList.add('d-none')
    if (iconMinimize) iconMinimize.classList.remove('d-none')
  })
  .catch(() => {
    // request denied — leave icons unchanged
  })
```

**Reason**
Prevents incorrect icon updates when fullscreen requests are denied or fail.

---

#### 2. Prevent Focus Errors in Keyboard Navigation

Added optional chaining when focusing the first element during focus trapping.

**Before**

```js
firstElement.focus()
```

**After**

```js
firstElement?.focus()
```

**Reason**
Avoids runtime errors when no focusable element is available.

---

#### 3. Ensure Error IDs Are Always Generated

Added a fallback generated ID when neither `input.id` nor `input.name` exists.

**Before**

```js
const errorId = `${input.id || input.name}-error`
```

**After**

```js
const errorId = `${input.id || input.name || accessibilityUtils.generateId('field')}-error`
```

**Reason**
Guarantees valid and unique error IDs for accessibility attributes.

### Testing

* Verified fullscreen icons update only after successful fullscreen entry.
* Verified no errors occur when focusable elements are missing.
* Verified error IDs are generated correctly for inputs without `id` or `name`.
* Ran project linting successfully.

### Type of Change

* Bug Fix
* Accessibility Improvement
* Code Quality Improvement

### DIFF
--- a/src/ts/accessibility.ts
+++ b/src/ts/accessibility.ts
@@ -351,7 +351,7 @@ export class AccessibilityManager {
   }
 
   private handleFormError(input: HTMLInputElement): void {
-    const errorId = `${input.id || input.name}-error`
+    const errorId = `${input.id || input.name || accessibilityUtils.generateId('field')}-error`
     let errorElement = document.getElementById(errorId)
     
     if (!errorElement) {
@@ -456,9 +456,10 @@ export class AccessibilityManager {
             event.preventDefault()
           }
         } else if (document.activeElement === lastElement) {
-          firstElement.focus()
+          firstElement?.focus()
           event.preventDefault()
         }
+
       }
     })
   }
--- a/src/ts/fullscreen.ts
+++ b/src/ts/fullscreen.ts
@@ -36,54 +36,64 @@ class FullScreen {
     this._config = config
   }
 
-  inFullScreen(): void {
+  async inFullScreen(): Promise<void> {
     const event = new Event(EVENT_MAXIMIZED)
 
     const iconMaximize = document.querySelector<HTMLElement>(SELECTOR_MAXIMIZE_ICON)
     const iconMinimize = document.querySelector<HTMLElement>(SELECTOR_MINIMIZE_ICON)
 
-    void document.documentElement.requestFullscreen()
+    try {
+      await document.documentElement.requestFullscreen()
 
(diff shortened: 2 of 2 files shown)


puikinsh:
Thanks for digging into these edge cases, @webgo-oss — really appreciated. 🙏

Your first point was a genuine bug: `void requestFullscreen()` flipped the icons and fired `maximized` even when the browser denied the request. That deserved fixing.

I've landed it in #6056 (now merged) with a slightly different approach. Rather than awaiting the request and updating the UI imperatively, a single **`fullscreenchange`** listener is now the source of truth for the icons and events. That fixes the denied-request case you found **and** a second desync the imperative version can't catch: exiting fullscreen with `ESC` or `F11` never runs `outFullscreen()`, so the icons used to get stuck. Keeping it event-driven also let the public methods stay `void`, matching their documented signatures in `fullscreen.mdx`.

On the other two changes:

- **Error-ID fallback** — right instinct (an input with no `id`/`name` collides on a literal `-error` id), but `generateId()` returns a new random value on every `invalid` event, so `getElementById` never matched and a fresh orphaned error `<div>` got appended on each re-validation. #6056 assigns the generated id to the element once so the same error node is reused.
- **`firstElement?.focus()`** — that branch only runs when `document.activeElement === lastElement`, which already guarantees the focusable list is non-empty and `firstElement` is defined, so I left it as-is.

Closing this in favour of #6056, but you're credited in the commit and the changelog. Thanks again for the report — please keep them coming!

Issue #5890, opened 2025-08-10
Question - Use AdminLTE4 with Hotwired Turbo


stehled:
I am currently trying to use AdminLTE v.4 in combination with Hotwired Turbo (https://github.com/hotwired/turbo). The idea is simple, instead of making full page reloads, swapout body or it's parts.

Unfortunately I am stuck on an issue, where I would need to reinitialize AdminLTE JS, so components like sidebar, treeview and others would work correctly. Normally, I would just listen to `turbo-reload` event and reinitialize need components. Unfortunately I was not able to find some global `init` function in AdminLTE, that I could call to reinitialize. How could I achieve this? I've found only this issue (https://github.com/ColorlibHQ/AdminLTE/issues/563) which wasn't that helpful.

Thank you

stehled:
Even though there are couple workarounds, they are not proper solutions and unfortunately AdminLTE is fixed to page loads too much. Closing as I've went with an alternative.

Issue #563, opened 2015-07-07
JQuery Turbolinks Issue using AdminLTE Theme


wonderer007:
I am using AdminLTE theme in my Rails project and have enabled Turbolinks, Everything is working fine except there is one issue that when i click on left menu my content-wrapper doesn't load properly instead i have to reload the page to load it properly. This is a very common issue in Turbolinks and we normally resolved by using 

$(function () { /* ... */ }); instead of $(document).on('ready', function () { /* ... */ });
and
$(document).on('click', 'button', function() { ... }) instead of $(document).on('click', 'button', function() { ... })  
So far i have tried in app.js to but no effect can you please guide me where do i need to change to load view properly .
Thanks

almasaeed2010:
I am not sure what TurboLinks is. Could you please clarify. 

Also, those are the same
> $(document).on('click', 'button', function() { ... }) instead of $(document).on('click', 'button', function() { ... })

Thanks!

billp:
I had an issue with the page height when I clicked on button with turbolinks enabled, the height of layout just shrinked. To solve it I just called $.AdminLTE.layout.activate(); from the CoffeeScript:

```
ready = ->
  $.AdminLTE.layout.activate();

$(document).ready(ready)
$(document).on('page:load', ready)
```

GCorbel:
I have the same problem. When I use a sidebar-toggle it works when I reload the page but not when I visit a new page and turbolink loads it.

The @billp 's code does not work. `$.AdminLTE.layout.activate()` do nothing.

billp:
Yes you are right, the  code works above in some cases only. 
Try this:

```
var o = $.AdminLTE.options
if (o.sidebarPushMenu) {
  $.AdminLTE.pushMenu.activate(o.sidebarToggleSelector) 
}
$.AdminLTE.layout.activate()
```

GCorbel:
This code works in coffeescript : 

    $(document).ready ->
      $.AdminLTE.layout.activate()

    $(document).on 'page:load', ->
      o = $.AdminLTE.options
      if o.sidebarPushMenu
        $.AdminLTE.pushMenu.activate(o.sidebarToggleSelector)
      $.AdminLTE.layout.activate()

And the equivalent in JavaScript : 

    $(document).ready(function() {
      $.AdminLTE.layout.activate();
    });

    $(document).on('page:load', function() {
      var o;
      o = $.AdminLTE.options;
      if (o.sidebarPushMenu) {
[...]

PR #6058, opened 2026-06-25
Resolved issue using hotwired/turbo links which broke PushMenu and TreeView

MarkDaleman:
Hello,

I was having the same problem as issue #5890 and #563  and since the newer Symfony 8 ships with hotwired/turbo I figured I wanted to solve this problem. Using the TreeView or PushMenu for the first time worked fine, but after clicking a link, it wouldn't work anymore. That's because hotwired/turbo replaces the <body> instead of navigating to a new page.

What I've done is creating a new state manager, which fires on the first load, as we're used to. However, it also hooks into turbo's specific lifecycle events (turbo:before-render and turbo:load). By resetting an internal initialization flag right before turbo renders the new content, the manager is able to re-trigger the initialization scripts needed for the TreeView and PushMenu to work.

### DIFF
--- a/.bundlewatch.config.json
+++ b/.bundlewatch.config.json
@@ -22,7 +22,7 @@
     },
     {
       "path": "./dist/js/adminlte.min.js",
-      "maxSize": "5.5 kB"
+      "maxSize": "5.56 kB"
     }
   ]
 }
--- a/src/ts/util/index.ts
+++ b/src/ts/util/index.ts
@@ -1,21 +1,29 @@
-const domContentLoadedCallbacks: Array<() => void> = []
+type DOMContentLoadedCallback = () => void;
 
-const onDOMContentLoaded = (callback: () => void): void => {
-  if (document.readyState === 'loading') {
-    // add listener on the first call when the document is in loading state
-    if (!domContentLoadedCallbacks.length) {
-      document.addEventListener('DOMContentLoaded', () => {
-        for (const callback of domContentLoadedCallbacks) {
-          callback()
-        }
-      })
+const adminLTEState = (() => {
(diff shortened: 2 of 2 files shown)


puikinsh:
Mark, thank you so much for this — genuinely. 🙏

Turbo support has been a long-standing pain point (#563, #5890) and honestly **we would not have tracked this down without your report**. You nailed the root cause — Turbo Drive replacing `<body>` so our one-time `DOMContentLoaded` init never re-runs — and your fix (reset a flag on `turbo:before-render`, re-init on `turbo:load`) was exactly the right idea. That diagnosis was the hard part.

I've shipped it in #6059 (now merged), with you credited as co-author on the commit. I reworked the implementation around one issue we found while reviewing: re-running **all** init callbacks also re-attaches the listeners that AdminLTE binds to `window`/`document` — the resize handlers, the accessibility keyboard/modal/dropdown listeners, the fullscreen `fullscreenchange` listener. Those targets *survive* Turbo's `<body>` swap, so re-running stacks a new copy of each on every navigation (an unbounded listener leak, plus duplicated screen-reader announcements over time).

So the merged version keeps your `turbo:load` / `turbo:before-render` lifecycle, but gives each init cycle an `AbortController`: the window/document listeners are registered with its signal, and the signal is aborted on `turbo:before-render` — tearing the old ones down before re-init. Listeners on `<body>` elements (TreeView/PushMenu buttons) need nothing, since Turbo discards the old body anyway. I also restored the immediate-call path for late/async registration and made the reduced-motion `<style>` inject only once.

Closing this in favour of #6059, but the credit is yours — this was a great catch. Please keep them coming, and thanks again for taking the time to diagnose it so clearly. 🙌

PR #6076, opened 2026-07-18
Fix push menu plugin to allow collapsed sidebar at initialization

dfsmania:
This pull request updates the sidebar initialization logic in `push-menu.ts` to better handle cases where the sidebar is explicitly set as collapsed. Now, the responsive logic for setting the sidebar state only runs if the sidebar is not already collapsed, preventing unwanted overrides. This way, a user can use the `sidebar-collapse` class on the body tag to get a collapsed sidebar on page initialization.

Sidebar state initialization improvements:

* Updated the sidebar initialization in the `PushMenu` class so that responsive logic is only applied if the sidebar is not already collapsed, ensuring explicit collapsed states are respected.

### DIFF
--- a/src/ts/push-menu.ts
+++ b/src/ts/push-menu.ts
@@ -341,12 +341,12 @@ class PushMenu extends BaseComponent {
 
     // When persistence is enabled and screen size is above the breakpoint, load
     // the saved sidebar state from local storage. Otherwise, use responsive
-    // logic to set the initial state. On low screen sizes, the sidebar should
-    // always be collapsed by default unless explicitly opened.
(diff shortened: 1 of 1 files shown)


puikinsh:
Thanks @dfsmania — clean fix. Verified locally before merging: ESLint on `push-menu.ts` is clean, the push-menu unit tests pass (4/4), and `tsc --noEmit` compiles. The `!isCollapsed()` guard correctly respects an explicit `sidebar-collapse` on `<body>` at init without regressing the mobile or mini-mode paths. Merging now 🎉

PR #6082, opened 2026-07-26
Add correct docstrings to base and push menu components.

dfsmania:
This pull request improves the code quality and maintainability of the `BaseComponent` and `PushMenu` classes by adding detailed documentation, clarifying lifecycle and registry management, and enhancing code readability.

**Documentation and Code Clarity Improvements:**

* Added comprehensive JSDoc-style comments and explanations to `BaseComponent`, its registry, and its lifecycle methods, making subclassing and usage clearer for future maintainers.
* Improved documentation for the `dispatchCustomEvent` utility, clarifying its parameters and return value.

**Component API and Lifecycle Enhancements:**

* Documented and clarified the API for `PushMenu`, including static methods like `getInstance` and `getOrCreateInstance`, and the constructor, making it easier to understand how to instantiate and interact with the component.

**Event Handling Consistency:**

* Refactored event dispatching in `PushMenu` methods (`expand` and `collapse`) for clarity: events are now stored in variables, and comments explain their cancelable behavior; added comments for "opened" and "collapsed" events.

**Code Formatting and Readability:**

* Reformatted long lines and query selectors for better readability, especially in DOM queries and media queries.

**Exports Organization:**

* Added explicit export blocks at the end of both files for clarity and maintainability.

### DIFF
--- a/src/ts/base-component.ts
+++ b/src/ts/base-component.ts
@@ -1,34 +1,61 @@
 /**
- * --------------------------------------------
+ * ----------------------------------------------------------------------------
  * @file AdminLTE base-component.ts
  * @description Shared component lifecycle for AdminLTE plugins: a per-element
  * instance registry (getInstance / getOrCreateInstance / dispose) and a
  * consistent custom-event contract, mirroring Bootstrap's component API.
  * @license MIT
- * --------------------------------------------
+ * ----------------------------------------------------------------------------
  */
 
 /**
- * element -> (data key -> component instance). WeakMap keys don't prevent
- * garbage collection, so instances die with their elements — important under
- * Hotwired Turbo, which swaps the whole <body> on navigation.
+ * Component registry:
+ * A WeakMap of elements to store component instances. Each element can have
+ * multiple component instances associated with it, keyed by the component's
+ * DATA_KEY: element -> (data key -> component instance).
+ * The WeakMap keys don't prevent garbage collection, so instances die with
+ * their elements, this is important under Hotwired Turbo, which swaps the
+ * whole <body> on navigation.
  */
 const componentRegistry = new WeakMap<Element, Map<string, BaseComponent>>()
 
+/**
+ * ----------------------------------------------------------------------------
+ * Class Definition
+ * ----------------------------------------------------------------------------
+ */
+
 class BaseComponent {
+  /**
+   * Get the component name. Subclasses must override this getter to provide
+   * their own name.
+   *
+   * @returns The component name.
+   */
   static get NAME(): string {
-    throw new Error('Component subclasses must override the static NAME getter.')
+    throw new Error('Component subclasses must override the NAME getter.')
   }
 
+  /**
+   * Get the component's data key.
+   *
+   * @returns The component's data key.
+   */
   static get DATA_KEY(): string {
     return `lte.${this.NAME}`
   }
 
   /**
+   * Get the component instance associated with the given element, if any.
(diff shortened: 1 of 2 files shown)


puikinsh:
Thanks @dfsmania — `BaseComponent` and `PushMenu` were the two files most in need of this, and several of the notes you added are exactly the kind a new subclass author needs.

I've pushed a follow-up commit to the branch (f73d605) rather than leaving a wall of line comments. It keeps the substance and drops the parts that were fighting existing conventions. The diff against `master` is now 53 lines, all comments, and the compiled bundle is byte-identical to `master` — I verified that by building both and diffing `dist/js/adminlte.js`.

### Kept

Everything that tells the reader something the signature doesn't: the registry semantics (multiple components per element, keyed by `DATA_KEY`), the `NAME` override contract, the `DATA_KEY` format, `_element`, the constructor's registration side effect, and full `@param`/`@returns` on `dispatchCustomEvent`. On `PushMenu`: `getInstance`/`getOrCreateInstance` — including the non-obvious bit that `config` is ignored when an instance already exists — plus `_config` and the constructor. I also carried over your idea of documenting `collapse()` symmetrically with `expand()`.

### Reverted, and why

**The line wrapping.** There's no `max-len` rule in `eslint.config.js`, and the codebase runs to 205 columns in `accessibility.ts` and 124 in `card-widget.ts`. Wrapping only these two files makes them look like a different codebase, and it adds conflict surface for no functional gain. The clearest symptom: the PR wrapped `getOrCreateInstance` in `push-menu.ts` while the byte-identical signature in `card-widget.ts` stayed on one line.

**The `const evt = ...` extraction.** `if (dispatchCustomEvent(...).defaultPrevented)` appears at seven call sites across `card-widget.ts`, `treeview.ts`, and `push-menu.ts`. Changing two of them leaves five inline and two extracted. (`evt` was also the only abbreviation of its kind in `src/ts` — `unicorn/prevent-abbreviations` is off, so lint won't catch it.)

**The `NAME` error message.** `'...must override the static NAME getter.'` → `'...must override the NAME getter.'` drops the word that tells a subclass author *which kind* of getter to write, which is the whole actionable content for someone who just wrote `get NAME()` as an instance method. Also worth flagging generally: that's a runtime string change inside a docs-only PR.

**The de-indented sample** in the Data Api docblock — the indent is what marked `PushMenu.getInstance(...)` as a code sample rather than prose.

**The banner changes.** Widening `base-component.ts`'s header to 76 dashes aligns it with `push-menu.ts` and `layout.ts` but diverges from the five files still on the short form, and the new `Exports` banners don't exist in any other module.

**Comments that restate their own code** — including one in `dispose()` that repeated the JSDoc three lines above it.

### Two follow-ups, if you're up for them

1. **A repo-wide pass.** `NAME`/`getInstance`/`getOrCreateInstance` are copy-pasted boilerplate in all seven components; documenting two of them is arbitrary. Same story for the banner styles, which are already inconsistent (`====` vs `----`, four different widths). Both are worth one deliberate sweep across `src/ts/` with an agreed convention, and I'd take that PR happily. Adding `eslint-plugin-jsdoc` in the same change would make the convention enforced rather than eyeballed.

2. **Sequencing.** Component lifecycle is still on the 4.2 list, and `BaseComponent`'s lifecycle is precisely what this documents. Might be worth landing this trimmed version now and holding the full sweep until that refactor settles, so the docs aren't written twice.

Merging once CI is green. Appreciate you picking at this — the accessor boilerplate genuinely was under-documented.

PR #6083, opened 2026-07-28
Fix AdminLTE for other SPA frameworks like GWT

themarioga:
This pull request improves AdminLTE's compatibility with GWT and other SPAs that have a race condition with AdminLTE after DOMContentLoaded is triggered but doesn't discard the body like Turbo.

For GWT users: you only need to call initAdminLTE after you handle current history.

```
	public static void initAdminLTE() {
		Scheduler.get().scheduleDeferred(AdminLTE::initAdminLte);
	}
	private static native void initAdminLte() /*-{
		$wnd.adminlte.initialize();
	}-*/;
```

### DIFF
--- a/src/ts/adminlte.ts
+++ b/src/ts/adminlte.ts
@@ -1,4 +1,4 @@
-import { onDOMContentLoaded } from './util/index.js'
+import { initialize, onDOMContentLoaded } from './util/index.js'
 import Layout from './layout.js'
 import CardWidget from './card-widget.js'
 import Treeview from './treeview.js'
@@ -40,5 +40,6 @@ export {
   FullScreen,
   PushMenu,
   ColorMode,
-  initAccessibility
+  initAccessibility,
+  initialize
 }
--- a/src/ts/push-menu.ts
+++ b/src/ts/push-menu.ts
@@ -469,16 +469,23 @@ onDOMContentLoaded(() => {
   // Handle touch events on overlay (area outside sidebar), usually we want to
   // close the sidebar when the user taps outside the sidebar on mobile
   // devices.
+  //
+  // These are bound with the lifecycle signal even though the overlay lives
+  // inside <body>: the node above is reused when it already exists, so under a
+  // framework that re-initialises against a persistent <body>, an unsignalled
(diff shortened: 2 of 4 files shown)


puikinsh:
Thanks @themarioga — this is a real gap and `initialize()` is the right shape for it. The delegated click handlers already survive anything, but the per-page init pass (PushMenu + overlay, treeview open-menu restore, color mode, a11y landmarks) had no public trigger, so frameworks that render into a persistent `<body>` after `DOMContentLoaded` were stuck faking Turbo events. Also good catch signalling the overlay listeners — under a persistent `<body>` those really do stack.

I verified everything locally (tests, lint, UMD/ESM builds — `window.adminlte.initialize()` is exported, so your GWT snippet works as written) and put your commit unchanged on [`feat/spa-initialize-api`](https://github.com/ColorlibHQ/AdminLTE/tree/feat/spa-initialize-api), with a follow-up commit (17fde86a) adding a few things the new API needs before it ships:

- **The same leak one layer down:** the accessibility module binds an unsignalled `invalid` listener on every form field each cycle, so two `initialize()` calls left three live handlers per input. Now registered with the lifecycle signal.
- **Early calls:** `initialize()` before the document finished loading marked the cycle initialised, and the real `DOMContentLoaded` pass then silently no-oped. The initial pass now runs through `initialize()` itself, so it always replays against the complete DOM.
- **Re-entrancy guard:** `initialize()` called from inside a lifecycle callback recursed forever.
- **`teardown()` export**, for apps that unmount the AdminLTE layout and want the cycle's `window`/`document` listeners gone without immediately rebinding them.
- Tests for all of the above (including an element-level listener case — the scenario your overlay fix covers), a FAQ entry, and a CHANGELOG credit.

One heads-up for your integration: components keep their instance per element, so on re-init against the *same* sidebar element `PushMenu.getOrCreateInstance()` returns the existing instance and changed `data-*` config attributes are not re-read — dispose first or use a fresh element if you need that.

I'll merge that branch — it contains your commit as-is, so this PR will show as merged — and it'll go out in the next release. Thanks again!

PR #6093, opened 2026-08-05
Improve color mode preference detection at initialization

dfsmania:
This pull request refactors and improves the `src/ts/color-mode.ts` module, enhancing clarity, maintainability, and theme selection logic. The most important changes are grouped below:

**Code organization and documentation improvements:**

* Improved and standardized section headers and comments throughout the file for better readability, and clarified the file-level documentation.

**Theme selection and validation enhancements:**

* Introduced a new `isValidTheme` method to centralize and clarify theme value validation.
* Added the `getDataBsTheme` method to support reading the theme from the `data-bs-theme` attribute, improving compatibility with server-side rendering and allowing theme selection via markup.
* Enhanced the theme selection logic in `getPreferredTheme` to check, in order: localStorage, the markup attribute, and finally OS preference, for a more robust and predictable user experience.

**Code clarity and maintainability:**

* Improved inline documentation and added comments to clarify logic in event listeners and initialization code, making the code easier to understand and maintain.
* Added a docstring to `_prefersDark` for consistency.

### DIFF
--- a/src/ts/color-mode.ts
+++ b/src/ts/color-mode.ts
@@ -1,23 +1,24 @@
 /**
- * --------------------------------------------
+ * ----------------------------------------------------------------------------
  * @file AdminLTE color-mode.ts
  * @description Color mode (light/dark/auto) switcher for AdminLTE.
- * Persists the choice in localStorage, follows the OS preference in
- * "auto" mode, and keeps [data-bs-theme-value] toggles and
- * [data-lte-theme-icon] indicator icons in sync.
+ * Persists the choice in localStorage, follows the OS preference in "auto"
+ * mode, and keeps [data-bs-theme-value] toggles and [data-lte-theme-icon]
+ * indicator icons in sync.
  *
- * Ships in the bundle so applications no longer need to copy the demo's
- * inline script. The tiny no-flash snippet in <head> (see _head.astro)
- * remains inline by design — it must run before first paint.
+ * Ships in the bundle so applications no longer need to copy the demo's inline
+ * script. The tiny no-flash snippet in <head> (see _head.astro) remains inline
+ * by design — it must run before first paint.
  * @license MIT
- * --------------------------------------------
+ * ----------------------------------------------------------------------------
  */
 
 import { getLifecycleSignal, onDOMContentLoaded } from './util/index'
 
 /**
+ * ----------------------------------------------------------------------------
  * Constants
- * ====================================================
+ * ----------------------------------------------------------------------------
  */
 
 const DATA_KEY = 'lte.color-mode'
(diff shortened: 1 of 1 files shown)


puikinsh:
Thank you for this, @dfsmania — genuinely. Honestly, we weren't aware ColorMode was stepping on `data-bs-theme` for anyone rendering a theme server-side; it only surfaced because you filed this alongside @cytech's #6084, and your diagnosis was exactly right: the module treated the markup as something to overwrite rather than something to read.

Your precedence chain is the core of what we've landed in #6095, unchanged — stored choice, then the theme the page declared, then the OS. While testing your branch, though, we hit two follow-on cases that the PR couldn't have seen from the source alone, so we extended it rather than merging as-is. Wanted to lay out the reasoning properly, since you'll run into the same thing next time you touch this module.

**1. The declared theme didn't survive an OS change.** Measured against your branch, built and run in headless Chromium with the OS preference starting light:

| `<html data-bs-theme="dark">`, nothing stored | after load | OS→dark | OS→light |
|---|---|---|---|
| 4.1.0 | `light` | `dark` | `light` |
| your branch | **`dark`** ✅ | `dark` | `light` ❌ |

`getPreferredTheme()` learned about the markup, but the `prefers-color-scheme` listener still only asked `getStoredTheme()` — so as far as it was concerned nothing had expressed a preference, and it was free to follow the OS. On a machine with a light/dark schedule, the app's theme flips at sunset and never comes back. The listener now consults the declared theme too.

**2. `data-bs-theme` is both the input and the output.** `getDataBsTheme()` reads at init, but by then — on any later pass (`turbo:load`, `initialize()`) — the value it reads is ColorMode's own previous write, not the page's intent; the two are indistinguishable after the first `_applyTheme()`. It's now captured once at module evaluation, which runs before any write under both lifecycle paths.

**3. The pre-paint snippet in `<head>` needed to come along.** This one was subtle and only turned up by running the demo: that snippet *always* writes `data-bs-theme` before the bundle loads, so once the module started treating the attribute as a declared preference, our own demo pages would have stopped following the OS — the snippet's output looks exactly like an authored theme. It now follows the same precedence (so it doesn't clobber a server-rendered theme before paint either) and marks values it worked out itself with `data-lte-theme-resolved`, which the module ignores.

**4. An explicit opt-out, for #6084.** Custom Bootstrap theme names (`data-bs-theme="myapp-dark"`) still fell through to the OS branch, and that's what @cytech is actually using — which is why they ended up forking `adminlte.esm.js` to stub `_applyTheme`. So there's now:

```html
<html data-lte-color-mode="off">
```

which stops every write — init, toggle clicks, OS changes — and is honoured by the inline snippet too.

Also added 13 unit tests over the precedence chain (`tests/unit/color-mode.test.ts` had no coverage of `getPreferredTheme()` at all), documented the chain and the opt-out on the Color Mode docs page, and removed the pre-4.1 inline toggler that page was still shipping — it bound its own handlers to the same `[data-bs-theme-value]` buttons and wrote to `theme` instead of `lte-theme`. One small thing we didn't carry over: the `====` → `----` banner restyle, only to keep `color-mode.ts` consistent with the other modules in `src/ts/`.

You're credited in the CHANGELOG and in the commit — the idea and the report are yours, and this is squarely a case of a contributor catching something the maintainers had missed. I'll close this one once #6095 merges, but please do keep them coming; your layout and push-menu work has been some of the most valuable input v4 has had. And if you spot a hole in *this* version, say so — clearly we benefit from it. 🙂

dfsmania:
Thanks for taking this into account! I'll test the new fix as soon as I can...

